PR #62, opened 2025-11-30
Set deterministic width for styling tests

max-sixty:
## Summary
- add a test-only terminal width override so styling logic can use a shared fixed width during tests
- update styling snapshot-related tests to rely on the shared width instead of per-call overrides and remove the unused width-specific formatter helper

## Testing
- cargo test --lib --bins

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_692b350bf3248325b5dcdf46193fa188)

### DIFF
--- a/src/styling/format.rs
+++ b/src/styling/format.rs
@@ -174,6 +174,15 @@ pub(super) fn wrap_styled_text(styled: &str, max_width: usize) -> Vec<String> {
 /// ```
 #[cfg(feature = "syntax-highlighting")]
 pub fn format_bash_with_gutter(content: &str, left_margin: &str) -> String {
+    format_bash_with_gutter_with_width(content, left_margin, get_terminal_width())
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks for identifying this issue! However, there's a simpler solution that doesn't require adding global state or `#[cfg(test)]` blocks.

The code already checks the `COLUMNS` environment variable at runtime. We just need to set it:

**Solution committed to `width-check` branch:**
1. `.cargo/config.toml` with `COLUMNS=80` - sets width for unit tests
2. Integration tests already override to `COLUMNS=150` in their command setup

This uses the existing `COLUMNS` support without adding:
- Global static variables
- `#[cfg(test)]` in production code  
- `Once` initialization patterns
- Per-test `set_test_columns()` calls

Closing in favor of the simpler approach.

PR #102, opened 2025-12-08
Improve command line completion on zsh

cquiroz:
First thanks for this tool, I have tested a few and this is the one most appropriate for my workflow

I noted that code completion is not fully working on zsh. This PR fixes  some edge cases that failed in main


max-sixty:
you're totally correct that the `zsh` autocomplete stopped working. apologies.

I'll merge this now and then add some tests, try to refine it

max-sixty:
will use #105 as it has tests, will credit you as the contributor @cquiroz

PR #105, opened 2025-12-08
Fix shell completion to bypass shell function

max-sixty:
## Summary

- Fix zsh lazy completion to use `command` builtin, bypassing the shell function
- Fix bash lazy completion with the same approach (same bug discovered during review)
- Add tests for both shells that verify the clap completer function gets defined

## Problem

When `WORKTRUNK_BIN` is not set, the lazy completion function calls `wt` which resolves to the shell function (not the binary). The shell function runs `wt_exec` which evals the completion script internally but doesn't re-emit it, causing the outer `eval` to receive nothing.

This manifested as:
```
_wt_lazy_complete:5: command not found: _clap_dynamic_completer_wt
```

## Solution

Use `command` builtin to bypass the shell function:
```bash
# Before (broken):
eval "$(COMPLETE=zsh "${WORKTRUNK_BIN:-wt}" 2>/dev/null)" || return

# After (fixed):
eval "$(COMPLETE=zsh command "${WORKTRUNK_BIN:-wt}" 2>/dev/null)" || return
```

## Credit

Thanks to (edit) @cquiroz for reporting this in #102!

## Test plan

- [x] Added `test_zsh_lazy_completion_defines_clap_function` 
- [x] Added `test_bash_lazy_completion_defines_clap_function`
- [x] Both tests verify the clap completer function is defined after triggering lazy completion
- [x] Tests simulate the real bug scenario (no `WORKTRUNK_BIN`, binary in `PATH`)

### DIFF
--- a/templates/zsh.zsh
+++ b/templates/zsh.zsh
@@ -54,7 +54,12 @@ if command -v {{ cmd_prefix }} >/dev/null 2>&1 || [[ -n "${WORKTRUNK_BIN:-}" ]];
             # Use `command` to bypass the shell function and call the binary directly.
             # Without this, `{{ cmd_prefix }}` would call the shell function which evals
             # the completion script internally but doesn't re-emit it.
-            eval "$(COMPLETE=zsh command "${WORKTRUNK_BIN:-{{ cmd_prefix }}}" 2>/dev/null)" || return
+            #
+            # The sed adds -V (unsorted group) and -o nosort to _describe, preserving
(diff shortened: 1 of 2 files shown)


jimmycuadra:
I'm confused why I'm mentioned here. I've never heard of this project...

max-sixty:
apologies @jimmycuadra , this must have been a mistake. I'll fix

Issue #101, opened 2025-12-07
Feedback and suggestions


pwntester:
Hi

Being using wt for the last week and loving it so far, thanks!

Here are a few ideas that I would find very useful:
- Generic hook configuration. I want to link/start/stop devcontainers when I switch to a worktree. This is something that i want to do always if .devcontainer/ exists so instead of having lots of .config/wt.toml around it would be nice to centrilize that logic in .config/worktrunk
- Add hook for post-remove so I can clean up devcontainers to `bar` worktree, I would like to create `../foo.bar` but then `pwntester/bar` branch with the `pwntester/` prefix

Thanks again for the project!

max-sixty:
thanks for the issue @pwntester !

> Add hook for post-remove so I can clean up devcontainers to `bar` worktree, I would like to create `../foo.bar` but then `pwntester/bar` branch with the `pwntester/` prefix

Great, will add. to confirm: this should run _post_ remove on the target (generally `main`) worktree? rather than pre-remove on the original worktree?

> Generic hook configuration. I want to link/start/stop devcontainers when I switch to a worktree. This is something that i want to do always if .devcontainer/ exists so instead of having lots of .config/wt.toml around it would be nice to centrilize that logic in .config/worktrunk

I thought about this...

...do you think folks really want a hook to run on _every_ project though? it's easy to add and wouldn't be _that much_ of an interface bloat. but I worry that it wouldn't get used much, wdyt?

we could add a `project_filter` field, but also someone can implement that in bash...

max-sixty:
I'm implementing a _pre_-remove hook, because it's much simpler given it's not in the background: https://github.com/max-sixty/worktrunk/pull/107. 

I think it mostly dominates the _post_-remove, because it can block the removal with a non-zero exit code, and there's no additional info we have post-remove...

pwntester:
Thanks for the pre-remove hook!
As for the generic hooks, they make a lot of sense to me since most of the actions will be shared across repos but I understand that may not be something universal

max-sixty:
great, check out #118

max-sixty:
@pwntester I'll close this as I think it's done, but please continue with more ideas!

PR #107, opened 2025-12-08
Add pre-remove hook for cleanup before worktree removal

max-sixty:
## Summary

- Add `pre-remove` hook that runs before `wt remove` deletes a worktree
- FailFast strategy: non-zero exit code aborts removal (worktree preserved)
- Works in both sync and background modes (hooks run before spawning background)
- Add `--no-verify` flag to skip hooks
- Add `wt step pre-remove` for manual execution

## Test plan

- [x] Hook executes with all template variables
- [x] Non-zero exit aborts removal
- [x] Works in background mode
- [x] Does not run for branch-only removal
- [x] `--no-verify` skips hooks
- [x] `wt step pre-remove` works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -490,9 +490,9 @@ dependencies = [
 
 [[package]]
 name = "criterion"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


max-sixty:
closes a subset of #101

PR #175, opened 2025-12-12
Enable merge readme tests on Windows

max-sixty:
## Summary
- Removed Windows skip from 3 merge tests that use mock shell scripts
- Git Bash handles shebang scripts without needing the execute bit
- The `chmod` was already conditionally compiled out with `#[cfg(unix)]`

**This is experimental** - if Windows CI fails, we'll know Git Bash does need something else.

## Test plan
- [x] Local macOS tests pass
- [ ] CI passes on Windows (the real test)


max-sixty:
Windows test failed with 'Failed to execute: git rev-parse --git-common-dir' - something about the test setup breaks git commands on Windows. This is a more complex issue than just chmod, so restoring the skip.

PR #235, opened 2025-12-16
Fix Windows race condition in multi-worktree removal

max-sixty:
## Summary

Fixes CI failure on Windows where `test_remove_multiple_including_current` was showing incorrect output for `feature-c`:

```
Old: 🔄 Removing feature-c worktree & branch in background (same commit as main)
New: 🔄 Removing feature-c worktree & branch in background
```

The integration reason was missing only for the second worktree in the sequence.

## Approach

Precompute integration reasons in `remove_worktree_by_name` before any output handling or spawning, rather than computing in `handle_removed_worktree_output` which happens after previous worktrees have spawned background processes.

**Note:** I'm not 100% certain this is the root cause. The failure pattern (only feature-c failing, not feature-a which processes after two backgrounds are running) doesn't perfectly fit a race condition theory. This PR tests whether precomputing fixes the issue.

## Changes

- Add `integration_outcome` field to `RemoveResult::RemovedWorktree`
- Compute integration check in `remove_worktree_by_name` and `remove_current_worktree`
- Use precomputed result in `handle_removed_worktree_output`
- Fail fast if `integration_outcome` not precomputed (programming error detection)

## Test plan

- [x] All 1508 tests pass locally
- [ ] CI passes on Windows (this is the real test)

### DIFF
--- a/src/commands/merge.rs
+++ b/src/commands/merge.rs
@@ -232,6 +232,16 @@ pub fn handle_merge(opts: MergeOptions<'_>) -> anyhow::Result<()> {
 
         // STEP 2: Remove worktree via shared remove output handler so final message matches wt remove
         let worktree_root = repo.worktree_root()?;
+        // After merge, the branch is integrated - compute integration reason for message
+        let destination_repo = Repository::at(&destination_path);
+        let effective_target = destination_repo.effective_integration_target(&target_branch);
+        let mut provider = worktrunk::git::LazyGitIntegration::new(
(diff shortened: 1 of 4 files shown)


max-sixty:
Closing - after further investigation, this was a one-off flaky test failure (1 failure among 20+ successful runs). Adding 79 lines of code to precompute integration reasons is excessive for a single sporadic failure. If this continues to fail, we should investigate more or add test-level retry logic.

PR #355, opened 2026-01-01
Fix PTY test to show compinit advisory

max-sixty:
## Summary

- Fixed the PTY test to use `WORKTRUNK_TEST_COMPINIT_MISSING=1` instead of `WORKTRUNK_TEST_COMPINIT_CONFIGURED=1`
- This skips the problematic `zsh -ic` probe (which leaks TTY output into PTY) while still testing the advisory output

The previous fix (#353) accidentally suppressed the compinit advisory warning by telling the code that compinit was already configured. This restores proper test coverage of the advisory.


max-sixty:
Closing - the diagnostic test was useful for investigation but doesn't belong in the test suite long-term.

Issue #372, opened 2026-01-02
Documentation incorrectly shows ~ concatenation with | filter without parentheses


coriocactus:
## Issue
Jinja2's operator precedence has pipe `|` with higher precedence than concatenation `~`, meaning expressions need parentheses to filter concatenated values: `'db-' ~ branch | hash_port` parses as `'db-' ~ (branch | hash_port)` therefore should be written as `('db-' ~ branch) | hash_port`. Similarly `repo ~ '-' ~ branch | hash_port` should be written as `(repo ~ '-' ~ branch) | hash_port`.

## Affected docs
https://worktrunk.dev/hook/#filters
https://worktrunk.dev/hook/#databases
https://worktrunk.dev/tips-patterns/#database-per-worktree

## Manual testing
```toml
[post-start]
echo = "echo {{ 'db-' ~ branch | hash_port }}"

# Output (invalid port):
# ◎ Running post-start project:echo:
# echo db-18979
```

```toml
[post-start]
echo = "echo {{ ('db-' ~ branch) | hash_port }}"

# Output (valid port):
# ◎ Running post-start project:echo:
# echo 11827
```

max-sixty:
thank you!

PR #373, opened 2026-01-02
fix: documentation incorrectly shows ~ concatenation with | filter without parentheses

coriocactus:
## Issue (https://github.com/max-sixty/worktrunk/issues/372)
Jinja2's operator precedence has pipe `|` with higher precedence than concatenation `~`, meaning expressions need parentheses to filter concatenated values: `'db-' ~ branch | hash_port` parses as `'db-' ~ (branch | hash_port)` therefore should be written as `('db-' ~ branch) | hash_port`. Similarly `repo ~ '-' ~ branch | hash_port` should be written as `(repo ~ '-' ~ branch) | hash_port`.

(FYI I'm a rust n00b. In addition to documentation changes, I've made code changes in `src/config/expansion.rs`.)

### DIFF
--- a/src/config/expansion.rs
+++ b/src/config/expansion.rs
@@ -295,9 +295,9 @@ mod tests {
 
         // Concatenation produces different (but deterministic) result
         let r1 = expand_template("{{ (repo ~ '-' ~ branch) | hash_port }}", &vars, false).unwrap();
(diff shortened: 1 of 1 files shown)


max-sixty:
thanks a lot @coriocactus !

PR #476, opened 2026-01-08
feat(switch): add --tmux flag to run worktree creation in tmux

lucasoeth:
## Summary

- Add `--tmux` flag to `wt switch` that runs worktree creation in a new tmux window/session
- Add `--detach` flag to run the tmux session in background instead of attaching
- All hooks (post-create, post-start, post-switch) run inside tmux
- Shell stays open at worktree path after completion

## Usage

```bash
# Create worktree in new tmux session (attached)
wt switch --create feature --tmux

# Create worktree in detached tmux session (background)
wt switch --create feature --tmux --detach
```

## Behavior

| Context | Default (`--tmux`) | With `--detach` |
|---------|-------------------|-----------------|
| Already in tmux | New window, switches to it | New window, switches to it |
| Not in tmux | New session, attaches | New session, detached |

## Test plan

- [x] `--tmux` conflicts with `--execute` (clap enforced)
- [x] `--detach` requires `--tmux` (clap enforced)
- [x] Help page snapshot updated
- [ ] Manual testing with tmux

### DIFF
--- a/src/commands/tmux.rs
+++ b/src/commands/tmux.rs
@@ -61,35 +61,39 @@ fn window_exists(name: &str) -> bool {
         .unwrap_or(false)
 }
 
-/// Collect environment variables to pass through to tmux session.
-/// Excludes internal/shell-specific vars that shouldn't be inherited.
-fn collect_env_exports() -> String {
-    let skip_vars = [
-        "PWD",
-        "OLDPWD",
-        "_",
-        "SHLVL",
-        "TMUX",
-        "TMUX_PANE",
-        "TERM_SESSION_ID",
-        "SHELL_SESSION_ID",
-        "WORKTRUNK_DIRECTIVE_FILE",
-        "COMP_LINE",
-        "COMP_POINT",
-    ];
+/// Environment variables to skip when passing to tmux session.
+const SKIP_ENV_VARS: &[&str] = &[
+    "PWD",
+    "OLDPWD",
+    "_",
+    "SHLVL",
+    "TMUX",
+    "TMUX_PANE",
+    "TERM_SESSION_ID",
+    "SHELL_SESSION_ID",
+    "WORKTRUNK_DIRECTIVE_FILE",
+    "COMP_LINE",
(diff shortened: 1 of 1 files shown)


max-sixty:
I'm open to something like this. Can we discuss the options?

Currently I use zellij, and my workflow is "new-tab", and then `wsl` expands to `wt switch --create --execute=claude`, and I add `-- 'Please fix foo'` to the end.

What's yours? What's the advantage of having "new-tab" be _within_ worktrunk?

lucasoeth:
The use case is agent-initiated worktree creation — Claude Code spawning parallel worktrees for handoffs that continue in the background. Your workflow works because you do the "new-tab" step manually, but when Claude wants to spawn a handoff there's no human to press new-tab first. It needs one atomic command that creates a detached session, creates the worktree inside it, runs the handoff, and continues in background.

That said, open to a more general approach. Could do something like `--detach-via "zellij run --"` where the user provides the wrapper command. That would be multiplexer-agnostic but puts more on the user. The tmux integration is what Claude Code needs today, but happy to generalize if there's a cleaner design.

max-sixty:
this is actually a very cool pattern, I haven't tried it much myself, it could be great

---

though I'm not sure it should be something _in_ Worktrunk... am I right that this could be:

```bash
tmux new-session -d -s fix-auth-bug "wt switch --create fix-auth-bug -x claude -- \
  'The login session expires after 5 minutes. Find the session timeout config and extend it to 24 hours.'"
```

...which is verbose, but tractable

could we start by having it as a Tip / Pattern at first? my preference would be to get more reps from folks running this sort of workflow before adding a feature to Worktrunk, unless it unlocks something significant we can't otherwise do — wdyt?

lucasoeth:
@max-sixty I like your suggestion, what about adding it as a skill that people can install?

max-sixty:
yes great!

@claude  can you draft a PR which adds this as a Tip & Pattern, and a section in our existing Claude Code Skill?

PR #480, opened 2026-01-08
fix(plugin): move hooks/skills to root to avoid EXDEV error

### LINKED ISSUE #478: Plugin fails to install on cross-filesystem setups (EXDEV error)
Your plugin fails to install on Ubuntu (and probably Fedora, Arch, etc).

The install command errors with:

```bash
$ claude plugin install worktrunk@worktrunk
Error: Failed to install: EXDEV: cross-device link not permitted, rename '/home/user/.claude/plugins/cache/worktrunk' -> '/tmp/claude-plugin-temp-1767905736814'
```

## Why this happens

You're using `"source": "./.claude-plugin"` in your marketplace.json. This triggers a bug in Claude Code's installer when /tmp and /home are on different filesystems (which is the default Ubuntu setup since 21.04).

I tested other plugins on my system and they install fine:

- perplexity uses `"source": "./"`
- claude-mem has no source field
- worktrunk uses `"source": "./.claude-plugin"` and fails

So it's specific to how your plugin is configured.

There's an upstream bug for this: anthropics/claude-code#14799

## Workaround for users right now

```bash
mkdir -p ~/.cache/tmp
TMPDIR=$HOME/.cache/tmp claude plugin install worktrunk@worktrunk
```

This works but users shouldn't have to do this.

## Fix options

**Quick fix:** Change `"source": "./.claude-plugin"` to `"source": "./"` in your marketplace.json and update the paths to be relative to the repo root instead of the .claude-plugin subdirectory.

**Bigger fix:** Move everything out of .claude-plugin/ to the repo root like most other plugins do.

Either way fixes the install issue. The first option is less invasive. And there's this issue with documentation for others until upstream makes their fix.

Once Claude Code fixes their installer bug (anthropics/claude-code#14799), you could keep using the subdirectory structure if you want. But that might take a while and users can't install the plugin in the meantime.

max-sixty:
## Summary

- Move `hooks/` and `skills/` from `.claude-plugin/` to repository root
- Change `source: "./.claude-plugin"` to `source: "./"` in marketplace.json
- Update documentation in `.claude-plugin/CLAUDE.md` explaining the structure

## Context

On Linux systems where `/tmp` is on a separate filesystem (Ubuntu 21.04+, Fedora, Arch), plugin installation fails with `EXDEV: cross-device link not permitted` when using `source: "./.claude-plugin"`.

See upstream issue: https://github.com/anthropics/claude-code/issues/14799

Fixes #478

## Test plan

- [x] `pre-commit run --all-files` passes
- [x] `cargo test --test integration test_skill_files_are_in_sync_with_docs` passes
- [ ] Plugin installation on Linux with tmpfs `/tmp`


max-sixty:
@rjmurillo lmk if you have any thoughts, thank you!

PR #498, opened 2026-01-09
feat: add sanitize_db filter for database-safe identifiers

### LINKED ISSUE #474: Add db_ident filter for database-safe identifiers
### Problem

The `sanitize` filter produces filesystem-safe strings but not database-safe identifiers. Branch names like `feature/auth-oauth2` become `feature-auth-oauth2`, which contains `-` — invalid in unquoted SQL identifiers.

```sql
CREATE DATABASE myapp_feature-auth-oauth2_dev;
-- ERROR: syntax error at or near "-"
```

### Proposed Solution

Add a `db_ident` filter that produces identifiers safe for unquoted use across SQL databases:

| Rule | Reason |
|------|--------|
| Lowercase | Portable across case-sensitive/insensitive databases |
| Replace non-alphanumeric with `_` | Only `[a-z0-9_]` universally safe |
| Collapse consecutive `_` | Cleaner output |
| Prefix `_` if starts with digit | Identifiers can't start with digits |
| Truncate to 63 chars | Fits all major databases (Postgres=63, MySQL=64, SQL Server=128) |

**Tested against**: PostgreSQL identifier requirements. Should work for MySQL, SQLite, SQL Server, and others following SQL standard identifier rules. Please open an issue if you find a database where this doesn't work.

**Transformations**:
| Input | Output |
|-------|--------|
| `feature/auth-oauth2` | `feature_auth_oauth2` |
| `123-bug-fix` | `_123_bug_fix` |
| `UPPERCASE.Branch` | `uppercase_branch` |

### Example usage

```toml
[post-create]
setup = """
createdb {{ repo }}_{{ branch | db_ident }}_dev
"""

[pre-remove]
cleanup = "dropdb {{ repo }}_{{ branch | db_ident }}_dev --if-exists"
```

### Current workaround

```bash
db_ident() { echo "$1" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9]/_/g; s/__*/_/g; s/_$//; s/^[0-9]/_&/' | cut -c1-63; }
```

max-sixty:
## Summary

- Add new `sanitize_db` template filter for database-safe identifiers
- Transforms branch names to `[a-z0-9_]` format safe for unquoted SQL identifiers
- Handles: lowercase conversion, underscore replacement, consecutive underscore collapse, leading digit prefix, 63-char truncation (PostgreSQL limit)

Example transformations:
| Input | Output |
|-------|--------|
| `feature/auth-oauth2` | `feature_auth_oauth2` |
| `123-bug-fix` | `_123_bug_fix` |
| `UPPERCASE.Branch` | `uppercase_branch` |

Closes #474

## Test plan

- [x] Unit tests for `sanitize_db` function covering all transformation rules
- [x] Unit tests for truncation behavior
- [x] Filter test via `expand_template`
- [x] Integration tests in `doc_templates.rs`
- [x] Non-ASCII character handling tests
- [x] All existing tests pass
- [x] Pre-commit lints pass

### DIFF
--- a/.claude-plugin/skills/worktrunk/reference/config.md
+++ b/.claude-plugin/skills/worktrunk/reference/config.md
@@ -166,7 +166,7 @@ The `[list]` section adds a URL column to `wt list`:
 url = "http://localhost:{{ branch | hash_port }}"
 ```
 
-URLs are dimmed when the port isn't listening. The template supports `{{ branch }}` with filters `hash_port` (port 10000-19999), `sanitize` (filesystem-safe), and `sanitize_db` (database-safe).
+URLs are dimmed when the port isn't listening.
 
 ### CI platform override
 
@@ -329,9 +329,10 @@ Creates `~/.config/worktrunk/config.toml` with the following content:
 
 # Worktree Path Template
 # Variables:
-#   {{ repo }}              - Repository directory name (e.g., "myproject")
-#   {{ branch }}            - Raw branch name (e.g., "feature/auth")
-#   {{ branch | sanitize }} - Branch name with / and \ replaced by - (e.g., "feature-auth")
+#   {{ repo }}                 - Repository directory name (e.g., "myproject")
+#   {{ branch }}               - Raw branch name (e.g., "feature/auth")
+#   {{ branch | sanitize }}    - Filesystem-safe: / and \ replaced by - (e.g., "feature-auth")
+#   {{ branch | sanitize_db }} - Database-safe: lowercase, underscores, hash suffix (e.g., "feature_auth_x7k")
 #
 # Paths are relative to the main worktree root (original repository directory).
 #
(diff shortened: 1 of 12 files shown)


max-sixty:
one issue is that it's not _that_ robust, for example: `/// - SQL reserved words (e.g., `user`, `select`) are not escaped`

- we could have a list of these and escape them, but then I worry we're maintaining a list of keywords...
- we could add something like a `_wt` suffix, or a suffix with the hash of the original string, which also would prevent collisions

hugobarauna:
Should this be mentioned [somewhere here](https://github.com/max-sixty/worktrunk/blob/b6f88689b20ee1c5899f64010834b5d51a80de72/docs/content/tips-patterns.md?plain=1#L83-L116) a well? 

Asking because that's what I read when I was setting up database per worktree for me.

max-sixty:
yes, good spot!

wdyt about these options? https://github.com/max-sixty/worktrunk/pull/498#issuecomment-3730423307

max-sixty:
Good catches on both points:

**Documentation gap**: Will update `docs/content/tips-patterns.md` to mention `sanitize_db` in the "Database per worktree" section.

**Reserved words / collisions**: The `_wt` suffix approach seems cleanest - it sidesteps both reserved words and collision issues without maintaining a keyword list. For example:
- `feature/auth` → `feature_auth_wt`
- `user` → `user_wt` (avoids reserved word)
- `a-b` and `a_b` → `a_b_wt` (still collides, but at least explicit)

The hash suffix idea (`feature_auth_a1b2`) would fully prevent collisions but makes identifiers less readable. Could offer both via a parameter or separate filter?

Happy to implement whichever direction you prefer.

max-sixty:
> wdyt about these options? [#498 (comment)](https://github.com/max-sixty/worktrunk/pull/498#issuecomment-3730423307)

I implemented them; will merge unless anyone has feedback

Issue #338, opened 2025-12-31
Add official Nix flake (package + Home Manager module)


Kabilan108:
## Description

Great work with this! Its solves a major pain point in my workflow with claude code and codex. 

I'd like to contribute a Nix flake for the project. I needed to add a flake to get it running on my nixos machines and I thought it would be helpful to share. I had codex add a flake.nix file that sets up the package and provides a home-manager module for configuration.

## Context

I can provide a flake with Rust toolchain pinning and completions generation ([fork here](https://github.com/Kabilan108/worktrunk/tree/main)). 
One open question: should tests run in the Nix build, or be disabled by default due to shell integration dependencies? I had trouble running the tests as part of the flake's build process.

Some background on nixos, flakes and home-manager:
- Flakes: https://nixos.wiki/wiki/Flakes
- Home Manager modules: https://nix-community.github.io/home-manager/options.html

max-sixty:
thanks @Kabilan108 !

could we submit to nixpkgs? iiuc that's the best place to put these?

soyboi:
> thanks [@Kabilan108](https://github.com/Kabilan108) !
> 
> could we submit to nixpkgs? iiuc that's the best place to put these?

No, this should be put in the project itself. Nix modules are what are put in nixpkgs, that is a different thing. Flakes are put in the projects themselves, and allow other nixos flakes to directly build it from the git repo

max-sixty:
Thanks for the contribution offer! I've been trying to understand the Nix ecosystem better. My current understanding:

**Arguments for nixpkgs + home-manager repos:**
- The [r-ryantm bot](https://nix-community.github.io/nixpkgs-update/r-ryantm/) automatically creates PRs for version bumps, so maintenance is handled by the ecosystem
- Standard installation path (`pkgs.worktrunk`)
- I don't use Nix myself, so I'd struggle to review/maintain Nix-specific code

**Arguments for an in-repo flake:**
- Users can install directly from any git ref (`nix build github:max-sixty/worktrunk/v0.9.1`) without waiting for nixpkgs to update
- Package and Home Manager module stay in sync with the code
- Per the [NixOS wiki](https://nixos.wiki/wiki/Flakes), flakes provide reproducibility and version locking

I'm genuinely uncertain which approach is better for users. @Kabilan108 @soyboi — what's your experience? Do most Nix users prefer pulling from nixpkgs, or do you find yourselves using in-repo flakes for tools you care about staying current on?

[...]

PR #502, opened 2026-01-09
feat: add Nix flake for packaging

marktoda:
Use crane for efficient Rust builds with dependency caching. 
Provides packages, dev shell, and CI checks (clippy, fmt, tests).

Fixes https://github.com/max-sixty/worktrunk/issues/338

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,5 @@ logo-variant-*.png
 test-results/
 
 # Nix build outputs
-result
-result-*
+/result
+/result-*


max-sixty:
great, thanks!

should we do a home-manager module? (I don't have a specific preference, that was just something discussed at the issue)

marktoda:
That's overkill at this point IMO, next step would be to upstream into nixpkgs once stabilized a bit. 

for now I think most nix users should just reference in flake.nix like
```
    worktrunk.url = "github:max-sixty/worktrunk";

```

craigdallimore:
Fantastic!

Is there a nixy way to set up the shell integration? I assume it's conventional for nix systems to define zshrc (etc) using nix.

```console
wt config shell install
❯ Install shell integration? [y/N/?] y

✗ Failed to configure zsh: Failed to open ~/.zshrc for writing: Permission denied (os error 13)
```

I guess we can add something like this in our system configuration - not sure if there is a better way. Apologies if this is the wrong place to discuss.

```
 if command -v wt >/dev/null 2>&1; then eval "$(command wt config shell init zsh)"; fi
```

max-sixty:
(I'm really not sure, totally fine to discuss it here though! I had thought that's maybe what Home Manager is for? but not sure)

Lewenhaupt:
> (I'm really not sure, totally fine to discuss it here though! I had thought that's maybe what Home Manager is for? but not sure)

Yeah that would typically be placed in home-manager. However, the eval route from the documentation isn't that hard for users to add themselves.

But as a reference for a program that adds config to the shell you can look at kitty which has an excellent home-manager module which shows exactly how this can be achieved:
https://github.com/nix-community/home-manager/blob/master/modules/programs/kitty.nix

max-sixty:
> Yeah that would typically be placed in home-manager. However, the eval route from the documentation isn't that hard for users to add themselves.

if anyone would like to do this, I would merge a high-quality PR (i.e. if you know nix and would own this being great, that would be v appreciated; please do not ask claude to one-shot it, particularly because I can't really assess its quality...)

PR #525, opened 2026-01-11
fix: remove apple_sdk framework

MattiasMTS:
Building this using nix-darwin causes failures as:
```
error: darwin.apple_sdk_11_0 has been removed as it was a legacy compatibility stub; see <https://nixos.org/manual/nixpkgs/stable/#sec-darwin-legacy-frameworks> for migration instructions
```

removing the `apple_sdk` solves it. Also took the opportunity to fix some linting issues ~and add the [nixfmt](https://github.com/NixOS/nixfmt) to the pre-commit~

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,8 +11,16 @@
     crane.url = "github:ipetkov/crane";
   };
 
-  outputs = { self, nixpkgs, flake-utils, rust-overlay, crane }:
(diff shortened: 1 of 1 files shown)


max-sixty:
thanks — I admittedly don't know this well, so will merge.

am fine to merge more if there are any other issues

PR #535, opened 2026-01-11
feat: add nushell shell support

MattiasMTS:
- Add Nushell variant to Shell enum with config paths and detection
- Create nushell.nu template with --env flag for cd directive handling
- Add nushell to default shells in install/uninstall commands
- Update completion format to support nushell descriptions
- Add nushell installation to CI workflows
- Document nushell setup in shell integration docs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,18 +54,24 @@ jobs:
     - name: 💰 Cache
       uses: Swatinem/rust-cache@v2
 
-    - name: Install shells (zsh, fish, nushell) - Ubuntu
+    - name: Install shells (zsh, fish) - Ubuntu
       if: runner.os == 'Linux'
       uses: awalsh128/cache-apt-pkgs-action@latest
       with:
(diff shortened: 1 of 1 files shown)


max-sixty:
very open to this, thanks for contributing!

we should ensure that it's well-tested, with some moderate share of the tests for existing shells. (we don't have to support all features, but we should be really confident that the ones we do support work well...)

I think it's acceptable to require `nushell` to be installed on a system in order for `--features=shell-integration-tests` to pass

MattiasMTS:
> very open to this, thanks for contributing!
> 
> we should ensure that it's well-tested, with some moderate share of the tests for existing shells. (we don't have to support all features, but we should be really confident that the ones we do support work well...)
> 
> I think it's acceptable to require `nushell` to be installed on a system in order for `--features=shell-integration-tests` to pass

Totally agree. I'll try to spend some time nailing this in later this week. Thanks for awesome maintenance and speed on this 🚤

max-sixty:
superseded, thanks for kicking it off though

PR #583, opened 2026-01-13
Do not open wt --help in pager by default

razor-x:
I'm not sure how intentional this was originally, but I have never used a CLI tool that forced --help to open in a pager. It's quite jarring and breaks expected workflows. Please consider merging this PR to bring this tool in line with standard expected CLI behavior for --help.

### DIFF
--- a/tests/integration_tests/shell_wrapper.rs
+++ b/tests/integration_tests/shell_wrapper.rs
@@ -3061,19 +3061,19 @@ echo "SCRIPT_COMPLETED"
         }
     }
 
-    /// Test that interactive `wt --help` uses a pager.
+    /// Test that interactive `wt --help` skips the pager.
     ///
     /// This is the complement to `test_wrapper_help_redirect_captures_all_output`:
     /// - Redirect case (`&>file`): pager should be SKIPPED (output goes to file)
-    /// - Interactive case (no redirect): pager should be USED
+    /// - Interactive case (no redirect): pager should be SKIPPED (output goes to terminal)
     ///
-    /// We verify pager invocation by setting GIT_PAGER to a script that creates
-    /// a marker file before passing through the content.
+    /// We verify pager is not invoked by setting GIT_PAGER to a script that would
+    /// create a marker file if it ran.
     #[rstest]
     #[case("bash")]
     #[case("zsh")]
     #[case("fish")]
-    fn test_wrapper_help_interactive_uses_pager(#[case] shell: &str, repo: TestRepo) {
+    fn test_wrapper_help_interactive_skips_pager(#[case] shell: &str, repo: TestRepo) {
         use std::io::Read;
 
         let wt_bin = get_cargo_bin("wt");
@@ -3194,12 +3194,12 @@ echo "SCRIPT_COMPLETED"
             terminal_output
         );
 
-        // Verify pager was invoked (marker file should exist)
+        // Verify pager was NOT invoked (marker file should not exist)
         assert!(
-            marker_file.exists(),
-            "{}: Pager was NOT invoked for interactive help.\n\
-             The marker file was not created, indicating show_help_in_pager() \n\
-             skipped the pager even though stderr is a TTY.\n\
+            !marker_file.exists(),
+            "{}: Pager was invoked for interactive help.\n\
+             The marker file was created, indicating show_help_in_pager() \n\
+             ran the pager even though help should print directly.\n\
              Terminal output:\n{}",
             shell,
             terminal_output


max-sixty:
this is what `git` does, though? or you see something different?

razor-x:
> this is what `git` does, though? or you see something different?

Negative, never seen that 🤔 Is it possible you have something configured on your shell that is hooking in and opening help in pagers? 

https://asciinema.org/a/j4QgsoGbqKVvCFv2

(Was going to look at updating the tests before moving out of draft)

max-sixty:
two things happening I think
- `wt` is opening a pager _even when there's space to display the content_. this doesn't happen on my end. I agree that's quite bad! thanks for raising...
- the `git` demo above doesn't test when the content is too big to fit — try `git diff --help` or reducing the size of the terminal — does it then happen?

razor-x:
Just checking the docs to see if it's something on my machine, but 

> git-help - Display help information about Git
>
> With no options and no `<command>` or `<doc>` given, the synopsis of the git command and a list of the most commonly used Git commands are printed on the standard output.
> https://git-scm.com/docs/git-help

Subcommands like `git log --help` shows the help for that command on stdout (which in this case prints the formatted man page to stdout)

But

> If a command or other documentation is given, the relevant manual page will be brought up.

So `git --help log` is like `man git log` (man viewer / pager)

image

razor-x:
> wt is opening a pager even when there's space to display the content. this doesn't happen on my end. I agree that's quite bad! thanks for raising...

Agree! `wt` is just an alias for `wt --help` right?

> the git demo above doesn't test when the content is too big to fit — try git diff --help or reducing the size of the terminal — does it then happen?

Yes you are right, `git diff --help` does open in a pager when the window is too small, but it's the man pager. 

But `git --help` does NOT open in a pager even when the window is too small, which is the typical behavior for other CLI tools.

Git sub command `--help` is all man pages instead of condensed usage, which is unusual but makes sense I guess for the scope of git.

max-sixty:
sorry, I didn't mean to close (and I definitely didn't mean to close the discussion!)

max-sixty:
> But `git --help` does NOT open in a pager even when the window is too small, which is the typical behavior for other CLI tools.

very viable to disable for `-h`, or for bare `wt`

I just put in #587 which should one of the possible causes of the pager even when there's enough space (which I don't see but is on your system, thanks for the demo)

---

_personally_ I very much like to have a pager when a long help opens, rather than scrolling back in my terminal. but my personal preferences should have a rapidly decreasing weight if this project is successful...

out of interest — would you _prefer_ no pager ever?

razor-x:
> _personally_ I very much like to have a pager when a long help opens, rather than scrolling back in my terminal. but my personal preferences should have a rapidly decreasing weight if this project is successful...
> 
> out of interest — would you _prefer_ no pager ever?

Calling out that we should compare to and match `git` earlier is a good objective approach. 😄 Matching what git and other long-standing CLI tools do by default should be the goal, that way, individuals can setup their environment to create their preferred behavior in a way that works consistently across all tools. (You can adjust your shell config to make this and all other tools always open in a pager if it follows never opening in one.)

To summarize git's behavior here

- Never opens in pager:
    - Command and subcommand usage: `git`, `git clone`, `git help`
    - Command and subcommand help flag: `git -h`, `git --help`, `git log -h`, `git log --help`
- Opens in pager because it's an alias to opening the help in the `man` pager:
    - `git --help git`, `git -h git`, `git --help log`, `git -h git`
    - `git help git`, `git help log`

For `wt`, this would map to

- Never opens in pager:
    - Command and subcommand usage: `wt`, `wt switch`
    - Command and subcommand help flag: `wt -h`, `wt --help`, `wt switch -h`, `wt switch --help`
- Opens in pager because it's an alias to opening the help in the `man` pager:
    - Not applicable because `wt --help list` is not implemented and does not use man page format.

> out of interest — would you _prefer_ no pager ever?

In terms of my personal preference, I like when the pager does not depend on the screen size, because I have muscle memory that assumes it always does the same thing. For example, on CLI, `foo --help` should never open a pager, and `man foo` should always open a pager. I also set in my git config to always use a pager for diff, since without it I would do things like run `git diff`, then `ctrl-d` to page down, and randomly exit my terminal on short or no diff output 😅 

```
[core]
	pager = delta --paging=always
```

max-sixty:
> since without it I would do things like run `git diff`, then `ctrl-d` to page down, and randomly exit my terminal on short or no diff output 😅

lol, ditto

---

(thank you for the very thorough and thoughtful response)

---

I would draw the distinction _slightly_ differently in terms of abiding by `git` — I'm not sure that in people's minds they are thinking that `git diff --help` "Opens in pager because it's an alias to opening the help in the man pager". Instead, it's that it's a long `--help` output, no?

Is there a model where we can still show it for long commands which you see as robust etc?

---

(the recent changes should at least now prevent showing help where there's enough space, which I know isn't your ideal, but appreciate you raising the issue and improving things)

razor-x:
> I would draw the distinction _slightly_ differently in terms of abiding by `git` — I'm not sure that in people's minds they are thinking that `git diff --help` "Opens in pager because it's an alias to opening the help in the man pager". Instead, it's that it's a long `--help` output, no?

Ok, wow, I was mistaken and thought `-h` and `--help` do the same thing, e.g., `-h` is just the short flag version of `--help`, but actually yes, `-h` is short output and `--help` is long (which they implemented as a man page). What's actually happening is that `--help` is just an alias for `git help` which is the man page format. I didn't test the difference between `-h` and `--help` enough before.

image

> 
> Is there a model where we can still show it for long commands which you see as robust etc?
> 
> (the recent changes should at least now prevent showing help where there's enough space, which I know isn't your ideal, but appreciate you raising the issue and improving things)

Yes, recent changes are an improvement. Based on the above, following git would mean 

- `-h` is short and never opens in a pager, so `wt`, `wt -h` and `wt switch -h` never open in a pager
- `wt --help <subcommand>` ("expands" to `wt help <subcommand>`) which would open in a pager (unclear if always or only if not enough room, as I can't find a short enough help page to test this 🤷🏻 ).
- Similarly, `wt <subcommand> --help` opens in a pager (alias for `wt help <subcommand>`)
- Special case though: `wt --help` and `wt help` never open in a pager since `git help` and `git --help` all seem be identical to `git -h`

Adding `wt help` is optional, but used as an example above to mirror what git is doing.

max-sixty:
yes I think that's correct!

practically I think that ends up very close to what we do now — the difference is whether `-h` opens a pager, but the terminal would have to be very small. And similarly for whether specifically `wt --help` opens a pager.

is that right? I could make that change. it makes the code a bit more complicated, because at the moment we just always use a pager, but only marginally, happy to do it

razor-x:
> yes I think that's correct!
> 
> practically I think that ends up very close to what we do now — the difference is whether `-h` opens a pager, but the terminal would have to be very small. And similarly for whether specifically `wt --help` opens a pager.
> 
> is that right? I could make that change. it makes the code a bit more complicated, because at the moment we just always use a pager, but only marginally, happy to do it

Yes, I do think at a minimum `wt` without any options that prints the usage/help should never open a pager, and same for any "bare" subcommands that print usage/error.

Beyond that I would just take the position that for consistency and principal of least surprise, `wt` follows git and `-h` never opens a pager  and `--help` does when needed. Kind of a long thread to get to this point, but since `help` is one of those things every user (especially first time users) will rely on, it feels important to get right 😅

max-sixty:
> Beyond that I would just take the position that for consistency and principal of least surprise, `wt` follows git and `-h` never opens a pager and `--help` does when needed. Kind of a long thread to get to this point, but since `help` is one of those things every user (especially first time users) will rely on, it feels important to get right 😅

yes, agree!

PR #561, opened 2026-01-12
Add Arch Linux install via AUR (btw)

razor-x:
- https://aur.archlinux.org/packages/worktrunk-bin
- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=worktrunk-bin

Followup ideally would be pushing a version bump via the release GitHub action. Happy to transfer ownership of the AUR package over for that.

PR #585, opened 2026-01-13
Add AUR publish workflow

max-sixty:
Adds automated AUR publishing to the release workflow, as suggested by @razor-x in #561.

## Changes
- Add PKGBUILD template in `.github/aur/`
- Add `publish-aur` job to release workflow using [github-actions-deploy-aur](https://github.com/KSXGitHub/github-actions-deploy-aur)
- Uses `updpkgsums` to auto-calculate checksums from release assets

## Setup required
- Generate SSH key and add public key to AUR account
- Add private key as `AUR_SSH_PRIVATE_KEY` GitHub secret

## Dependency
Requires #561 to be merged first (adds Arch Linux install instructions to docs).

cc @razor-x

### DIFF
--- a/.github/aur/PKGBUILD
+++ b/.github/aur/PKGBUILD
@@ -16,6 +16,12 @@ sha256sums_aarch64=('SKIP')
 package() {
   cd "worktrunk-$CARCH-unknown-linux-musl"
   install -Dm755 wt "$pkgdir/usr/bin/wt"
-  install -Dm755 git-wt "$pkgdir/usr/bin/git-wt"
   install -Dm644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+
+  # Fish completions
+  # TODO: a bit hacky, but short enough that copy & paste is OK for the moment
+  install -Dm644 /dev/stdin "$pkgdir/usr/share/fish/vendor_completions.d/wt.fish" <<'EOF'
+# worktrunk completions for fish
+complete --keep-order --exclusive --command wt --arguments "(test -n \"\$WORKTRUNK_BIN\"; or set -l WORKTRUNK_BIN (type -P wt 2>/dev/null); and COMPLETE=fish \$WORKTRUNK_BIN -- (commandline --current-process --tokenize --cut-at-cursor) (commandline --current-token))"
+EOF
 }


.github/aur/PKGBUILD
```diff
@@ -0,0 +1,21 @@
+# Maintainer: Maximilian Roos <m@maxroos.com>
+pkgname=worktrunk-bin
+pkgver=0.0.0
+pkgrel=1
+pkgdesc="CLI for git worktree management, designed for running AI agents in parallel"
+arch=('x86_64' 'aarch64')
+url="https://github.com/max-sixty/worktrunk"
+license=('MIT' 'Apache-2.0')
```
razor-x: I had to look this up before: 

```suggestion
license=('MIT OR Apache-2.0')
```

> Combining multiple licenses or adding exceptions should follow the [SPDX syntax](https://spdx.github.io/spdx-spec/v3.0/annexes/SPDX-license-expressions/). For example a package released under either GNU/GPL 2.0 or GNU/LGPL 2.1 could use 'GPL-2.0-or-later OR LGPL-2.1-or-later', a package released under Apache 2.0 with LLVM exception would use 'Apache-2.0 WITH LLVM-exception' and a package released with part under the BSD 3 clause, others under GNU/LGPL 2.1 and some under GNU/GPL 2.0 would use 'BSD-3-Clause AND LGPL-2.1-or-later AND GPL-2.0-or-later'[[2]](https://gitlab.archlinux.org/archlinux/packaging/packages/musepack/-/blob/f9642f78a1a7c61e6b8d181b138ee3f7e8717054/PKGBUILD#L16). **Note that this must be a single string, so the entire expression has to be enclosed in quotes.** As for November 2023 [SPDX list of exceptions](https://spdx.org/licenses/preview/exceptions-index.html) is limited, so usually the custom license route must be used.
>
> https://wiki.archlinux.org/title/PKGBUILD#license

.github/aur/PKGBUILD
```diff
@@ -0,0 +1,21 @@
...
+conflicts=('worktrunk')
+source_x86_64=("https://github.com/max-sixty/worktrunk/releases/download/v${pkgver}/worktrunk-x86_64-unknown-linux-musl.tar.xz")
+source_aarch64=("https://github.com/max-sixty/worktrunk/releases/download/v${pkgver}/worktrunk-aarch64-unknown-linux-musl.tar.xz")
+sha256sums_x86_64=('SKIP')
+sha256sums_aarch64=('SKIP')
+
+package() {
+  cd "worktrunk-$CARCH-unknown-linux-musl"
+  install -Dm755 wt "$pkgdir/usr/bin/wt"
+  install -Dm755 git-wt "$pkgdir/usr/bin/git-wt"
```
razor-x: I didn't really understand why both binaries would want to be installed (what is `git-wt` for?). Would this be needed on an Arch Linux system?

.github/aur/PKGBUILD
```diff
@@ -0,0 +1,21 @@
...
+  install -Dm644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+}
```
razor-x, line 27: Installing the fish completions to `${pkgdir}/usr/share/fish/vendor_completions.d/wt.fish` is missing here but standard for Arch Packages. 

Since there is no `wt config shell completions fish` command I had to "fake the install" (without the env vars, the build step will clobber the actual home directory)

```sh
build() {
  mkdir -p "home/.config/fish/conf.d/" \
    "home/.config/fish/conf.d/" \
    "home/.config/fish/completions/" \
    "home/.config/fish/functions/"

  HOME="${srcdir}/home" \
    ZDOTDIR="${srcdir}/home" \
    XDG_CONFIG_HOME="${srcdir}/home/.config" \
    XDG_DATA_HOME="${srcdir}/home/.local/share" \
    "./worktrunk-$CARCH-unknown-linux-musl/wt" config shell install -y
}

package() {
  // ...
  mkdir -p "${pkgdir}/usr/share/fish/vendor_completions.d/"
  install -Dm644 "../home/.config/fish/completions/wt.fish" "${pkgdir}/usr/share/fish/vendor_completions.d/wt.fish"
}
```

max-sixty: I added this manually, but I'm actually not sure it's really necessary — it's so important to install the shell integration that it's very unlikely that someone will be missing the completions...

razor-x: Yes, installing the completions is expected and shows good faith to serve the users of the platform. The completions could be PRed directly to the fish GitHub repo and then this step would be unnecessary. Fish likes to have completions shipped with the shell. However, until that happens, it's important that they're available from the package install.


In terms of the shell integration, that is provided separatly from the completions, which is correct for fish.

I manage all my dotfiles with a dotfile manager and fisher. Tools that only provide an "install" path that clobbers dotfiles without any escape hatch are incompatible with idopotent tools. The inti command is compatible because it prints to standard out which I can hook into such tooling. The worktrunk is written right, now I would be forced to run the install command to "get" the completions which would require a lot of workarounds added to my config management.

razor-x: For reference: https://github.com/razor-x/dotfiles/blob/3c462f4f3c3e23db943d9b2b3f53d034dbdd4386/home/dot_local/share/exact_plugin/exact_fish/exact_functions/wt.fish

razor-x (approved):
Let's give it try! I'm getting tired of updating this by hand 😅 Once confirmed working, please merge  #561 I'll abandon the AUR package.

I tested the PKGBUILD locally and it worked!

PR #648, opened 2026-01-15
Add Winget installation instructions to README

Farley-Chen:
Added installation instructions for Winget on Windows.

Cargo install for windows will occur error because of two reason.

1. the script in `Microsoft.PowerShell_profile.ps1` is `if (Get-Command wt -ErrorAction SilentlyContinue) { Invoke-Expression (& wt config shell init powershell) }` will cause `Cannot convert 'System.Object[]' to the type 'System.String`

2. the init script `Get-Command wt -CommandType Application` will cause `The term 'wt.exe wt.exe' is not recognized as a name of a cmdlet, function, script file, or executable program.` because the exists `wt.exe` for windows

I think the best way for windows install is to use winget and use `git-wt` command

### DIFF
--- a/.claude-plugin/CLAUDE.md
+++ b/.claude-plugin/CLAUDE.md
@@ -42,4 +42,4 @@ The hooks track Claude Code activity via git config (`worktrunk.status.{branch}`
 
 **Problem**: If the user interrupts Claude Code (Escape/Ctrl+C), the 🤖 status persists because there's no `UserInterrupt` hook. The `Stop` hook explicitly does not fire on user interrupt.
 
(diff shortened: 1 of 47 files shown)


max-sixty:
Thanks for the detailed bug report! I've opened #652 to fix issue (1) — the PowerShell script now handles the case where multiple `wt.exe` binaries exist.

Your Winget installation instructions are still valuable for users who prefer to avoid the name collision entirely by using `git-wt`.

max-sixty:
I placed it differently but gave you credit on the release notes. thank you!

PR #655, opened 2026-01-15
Add [switch] create config option

JoshuaKirby88:
You can now set `[switch] create = true` in your config to make `wt switch <branch>` behave like `wt switch --create <branch>`


max-sixty:
> The [switch] create = true config option was documented but never implemented.

what's the source for this claim?

JoshuaKirby88:
Sorry, I thought I remembered seeing this option.
Thanks for checking.

PR #698, opened 2026-01-17
(wip, draft): add git-compatible ref path parsing for branch resolution

max-sixty:
## Summary

Add git-compatible `refs/heads/X` and `heads/X` syntax for branch resolution, providing an escape hatch for special branch names like `@`.

Everything else passes through unchanged — git handles it naturally.

## Test plan

- [x] `wt switch refs/heads/feature` resolves to branch `feature`
- [x] `wt switch heads/feature` resolves to branch `feature`  
- [x] `wt switch refs/heads/@` escapes the special `@` symbol
- [x] `wt switch refs/tags/v1.0` passes through, git shows "branch not found"
- [x] Empty prefixes like `refs/heads/` give clear error

_This was written by Claude Code on behalf of max-sixty_


max-sixty:
Closing after analysis determined this refactoring is not worth the cost.

## Why the current approach is better

1. **Worktrunk only operates on local branches** - Git worktrees can only be on local branches (`refs/heads/*`). There's no case where `foo` is ambiguous.

2. **Current approach has fewer conversions** - Storing full refs would require stripping for display, config, and comparisons. Current approach only needs stripping on parse (2 places).

3. **The escape hatch already works** - The minimal fix in this PR (accepting `refs/heads/@` to refer to branch `@`) handles the edge case without internal representation changes.

4. **Config migration burden** - Existing configs use short names (`worktrunk.history`, `worktrunk.default-branch`, etc.). Migration would require version detection and potentially break user scripts.

The git-compatible ref path parsing from this PR was already merged separately in 94ab866db, providing the necessary escape hatch for accessing branches named `@`, `-`, or `^`.

> _This was written by Claude Code on behalf of maximilian_

PR #701, opened 2026-01-17
feat: add completion cmd for completion integration

chenrui333:
- https://github.com/Homebrew/homebrew-core/pull/262653

cc @max-sixty


max-sixty:
thanks a lot @chenrui333 . I made a change to move it below the top level, because I'm very keen to keep the top level very clean. I updated https://github.com/Homebrew/homebrew-core/pull/262653/changes too.

Hope that's OK

PR #716, opened 2026-01-18
chore: update brew install to use core tap

chenrui333:
Update brew install to use core tap per the pr merge, https://github.com/Homebrew/homebrew-core/pull/262653


max-sixty:
thank you @chenrui333 ! I will add back the `&& ...`, because that's still required for the shell extension (otherwise we can't change dirs on `wt switch`)

PR #762, opened 2026-01-20
fix: use --show-toplevel for repo_path() to support submodules

lajarre:
The previous implementation derived repo_path from git_common_dir's parent, which fails for submodules where git data is stored in the parent repo's .git/modules/ directory.

## Bug report (previous implementation)

Problem: When running wt switch --create <branch> inside a git submodule, worktrees are created in the wrong location.

Reproduction:

```shell
# Inside a git submodule at /parent/submodule
wt switch --create feature

# Expected: /parent/submodule/.worktrees/feature
# Actual:   /parent/.git/modules/.worktrees/feature
```

Note: This affects configurations with relative worktree-path templates like `.worktrees/{{ branch | sanitize }} or .worktrees/{{ branch | sanitize }}`. The default sibling-directory config (`../{{ repo }}.{{ branch | sanitize }}`) may also be affected since it's relative to the incorrectly computed repo root.

Root Cause: `repo_path()` derived the working directory from `git_common_dir.parent()`, assuming .git is always inside the working tree. For submodules, git stores data in `parent/.git/modules/sub`, so `parent()` returns the wrong path.

Fix: Use `git rev-parse --show-toplevel` which correctly returns the working tree root regardless of where git stores its data.

### DIFF
--- a/src/git/repository/mod.rs
+++ b/src/git/repository/mod.rs
@@ -314,44 +314,60 @@ impl Repository {
         self.git_common_dir().join("wt-logs")
     }
 
-    /// The repository root path.
+    /// The repository root path (the main worktree directory).
     ///
-    /// For normal repositories: the main worktree directory (parent of .git).
-    /// For bare repositories: the bare repository directory itself.
+    /// - Normal repositories: the main worktree directory (parent of .git)
+    /// - Bare repositories: the bare repository directory itself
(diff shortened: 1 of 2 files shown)


max-sixty:
thank you @lajarre !

I made a small change to avoid looking at the paths. does that look OK to you?

lajarre:
> I made a small change to avoid looking at the paths. does that look OK to you?

Much better, lgtm!

max-sixty:
thanks a lot @lajarre

PR #776, opened 2026-01-21
docs: add worktrunk Agent Skills file

jaydenfyi:
Add an [Agent Skills](https://agentskills.io/home) file to better allow AI agents to create and manage their own worktrees.

Would allow people to call
```bash
npx skills add max-sixty/worktrunk
```

for people to easily add the skill to their own skills directories.

Just for consideration. I can see how this comes with extra/duplicate documentation overhead as well.

### DIFF
--- a/skills/worktrunk/SKILL.md
+++ b/skills/worktrunk/SKILL.md
@@ -1,0 +1,250 @@
+---
+name: worktrunk
+description: Manage git worktrees for parallel development workflows. Use when working with git worktrees, feature branch workflows, parallel AI agent tasks, or when the user mentions wt, worktrees, branch switching, or merging feature branches.
+allowed-tools: Bash(wt:*)
+---
+
+# Worktrunk (wt)
+
+## Quick start
+
+```bash
+wt switch --create feature    # Create worktree and branch
+wt switch feature             # Switch to worktree
+wt list                       # Show all worktrees
+wt merge                      # Squash, rebase, merge to main, remove worktree
+wt remove                     # Remove worktree; delete branch if merged
+wt config shell install       # Install shell integration (required for cd)
+```
+
+## Core concepts
+
+**Worktrees** are separate working directories per branch (unlike `git switch` which changes branches in place).
+**Branches** are addressed by name; paths are computed from templates.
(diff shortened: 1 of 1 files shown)


max-sixty:
is `npx skills` a popular deployment mechanism?

we currently have a skill in the Claude Code plugin; keen to allow other agents to use it too...

jaydenfyi:
> is `npx skills` a popular deployment mechanism?
> 
> we currently have a skill in the Claude Code plugin; keen to allow other agents to use it too...

Ah apologies, I was unaware the Claude plugin already had a skill file defined 🤦‍♂️

Definitely _very_ early, I've only just seen some mentions of it in my Twitter bubble over the last few days. Here's a blog post from 12 hours ago 😅 https://vercel.com/changelog/introducing-skills-the-open-agent-skills-ecosystem

But I would bet on it gaining adoption based on it being interoperable with any coding agent, and by virtue of triangle companies influence. I don't personally use Claude Code which is why it's only recently caught my attention as Codex and other agents are starting to add support for skills.

Happy for you to close this if it's not desirable yet or not documented how you'd like. I think I aimed for it to be pretty specific to using the worktrunk CLI commands directly, which I guess is quite a bit different to the existing one in the Claude plugin.

max-sixty:
ok, let's keep an eye on how things evolve on the non-claude-code approach to installing skills?

max-sixty:
I took a look at the `npx skills` — it looks decent!

I can't seem to work out how to tell it where our skills are! we don't want our `release` skill installing for others (that's how to release worktrunk...), which is the only one it finds

we could merge https://github.com/max-sixty/worktrunk/pull/480, which might help it find the actual worktrunk skill, but we still have the `release` issue

if we can make unobtrusive changes to the repo and become compatible, that would be great!

jaydenfyi:
An update: the skills CLI now respects an `internal: true` metadata tag that could be used to exclude the release skill.

https://github.com/vercel-labs/skills#optional-fields

max-sixty:
I added the `internal: true` a while ago, sorry for not reporting back here. Thank you for the issue! I'll close this but any other ideas about the skill: lmk

Issue #738, opened 2026-01-19
Question: Is there a command to remove a worktree and checkout the branch in main?


zpeleg:
Hello,

I am looking for a way to streamline my workflow. Often, I take a specific branch from a worktree and want to run it from the main git folder immediately after removing the worktree.

Currently, I am doing this:
```
git status # to find branch name
worktrunk remove
# (automatically moved to main git folder)
git checkout <copied_branch>
```

Is there an existing command or flag in worktrunk that can do this in one step?

Thank you :)

max-sixty:
Thanks a lot for kicking off the issue

this is something that folks have asked for before. I'm very hesitant to add it, because:
- it puts the repo in a "mismatched branch/worktree state" — i.e. the branches aren't in the worktrees where we expect them to be (denoted by the red flags in `wt list`)
- we no longer have a worktree with `main` checked out, which is generally helpful to have around
- I generally would have thought that the idiomatic workflow — of creating short-lived worktrees in the correctly mapped locations — would be sufficient

...but I also don't want to be obstinate or force a very narrow workflow, especially as more people use Worktrunk

Would it be possible to describe some use-cases where this would be particularly helpful? for example there was a good case from @joelgrus where testing OAuth is restricted to certain ports such that it's not possible to just spin up more worktrees on different ports

any others?

ziv-nagish:
I definitely understand the desire to keep a clean state, but for many workflows, the project root has a certain "gravity" due to how external tools are configured. Here are two specific use cases where this command would be a lifesaver:

* **Tooling & Orchestration (e.g., Docker Compose)**: Many projects are configured to run microservices via Docker Compose or scripts that rely on relative paths or environment files strictly tied to the root directory. Reconfiguring these to work within a worktree subdirectory is often a high-effort task that's just not worth it compared to simply moving the branch to the root.

* **Heavy Build Artifacts (e.g., React Native/Mobile)**: Frameworks like React Native generate local artifacts, build caches, and native IDE configurations (Xcode/Android Studio) that are anchored to the project root. Re-initializing these within a temporary worktree is often impractical or "impossible" due to the sheer size and time required for a fresh build.

In these cases, I use the worktree to code and review, but I need the root directory to actually run and test. Being able to "promote" a branch from a worktree to the root in one step would significantly reduce the friction of this workflow.

[...]

PR #789, opened 2026-01-22
Add experimental `wt step promote` command

max-sixty:
## Summary

Adds `wt step promote` — an experimental command to exchange branches between the main worktree and another worktree, including swapping gitignored files (build artifacts, `.env`, `node_modules/`, etc.). This addresses #738 where users want to move a branch from a worktree into the main worktree.

This is useful when the main worktree has special significance (Docker Compose configs, IDE settings, heavy build artifacts anchored to project root) and hooks aren't yet set up to run on arbitrary worktrees.

**Key features:**
- Exchange branches between main worktree and any other worktree
- Swap gitignored files so build artifacts follow their branch
- Auto-restore: `wt step promote` from main worktree (no args) restores default branch
- Shows mismatch state in `wt list` with ⚑ indicator
- Warns when creating mismatch, with hint to restore
- `.worktreeinclude` filtering respected per worktree

**Swap algorithm (stage → exchange → distribute):**
1. Discover gitignored entries before branch exchange (`.gitignore` belongs to the branch)
2. Stage files to `.git/wt-promote-staging/` before `git switch` (prevents tracked file collision)
3. Exchange branches with per-step rollback
4. Cross-distribute staged files to new worktrees
5. Best-effort restore on failure with recovery guidance

**Adversarial testing (4 rounds, 9 issues found and fixed):**
- Stage before switch to prevent tracked file collision
- Per-step rollback in `exchange_branches()`
- Bail on leftover staging (don't destroy previous run's files)
- Recovery guidance on all failure paths
- Dedicated `restore_staged()` for rollback semantics
- `ErrorKind::CrossesDevices` for cross-device fallback
- `Option<PathBuf>` instead of sentinel empty path

## Test plan

- [x] 25 integration tests covering: bidirectional swap, round-trip, `.worktreeinclude` filtering, nested directories, same-filename-different-content, tracked-vs-ignored separation, stale staging detection, bare repo rejection, detached HEAD, dirty worktrees, no-arg from main/linked, branch argument, already-in-main, mismatch indicator in `wt list`
- [x] All 2576 tests pass
- [x] All lints clean
- [x] `copy-ignored` tests still pass after `list_and_filter_ignored_entries` extraction

### DIFF
--- a/src/commands/step_commands.rs
+++ b/src/commands/step_commands.rs
@@ -1018,8 +1018,8 @@ fn exchange_branches(
     target_branch: &str,
 ) -> anyhow::Result<()> {
     let steps: &[(&worktrunk::git::WorkingTree<'_>, &[&str], &str)] = &[
-        (target_wt, &["checkout", "--detach"], "detach target"),
-        (main_wt, &["checkout", "--detach"], "detach main"),
+        (target_wt, &["switch", "--detach"], "detach target"),
+        (main_wt, &["switch", "--detach"], "detach main"),
         (main_wt, &["switch", target_branch], "switch main"),
         (target_wt, &["switch", main_branch], "switch target"),
     ];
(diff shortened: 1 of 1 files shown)


max-sixty:
**Limitation: gitignored files don't move with the branch**

The current `wt step promote` only swaps branches — gitignored/untracked files (`target/`, `node_modules/`, `.env`) stay in their original filesystem locations. A developer who runs expensive builds in a linked worktree loses their build artifacts when promoting.

**Why this is tricky:**

- **`git worktree move` doesn't work for the main worktree** — Git doesn't support it. This is the same limitation `relocate` hits; for main worktree relocation, it checks out the default branch and creates a new linked worktree elsewhere (files don't move).

- **Filesystem swap + metadata repair is risky** — We could swap directories and fix `.git/worktrees/*/gitdir` pointers, but the main worktree has an embedded `.git/` directory while linked worktrees have a `.git` file. Lock files, hooks, and index state could get confused. High risk of corruption.

- **Copy-based swap has tradeoffs** — Using the `copy-ignored` infrastructure (reflink where available) avoids touching git metadata, but: not atomic (power failure mid-swap loses files), needs 2x disk space during swap, slow for large directories even with reflink.

- **Accepting the limitation** — Document that `promote` only swaps branches. Simple and safe, but users must manually copy files or rebuild.

---

_This comment was written by Claude Code on behalf of @max-sixty_

PR #790, opened 2026-01-22
Add `wt step relocate` command

max-sixty:
## Summary

- Add `wt step relocate` command to move worktrees to their expected paths based on the `worktree-path` template
- Handles swap/cycle scenarios automatically via temporary locations
- Add `--clobber` flag to backup non-worktree paths at target locations
- Add `--commit` flag to auto-commit dirty worktrees before relocating

## Key behaviors

**Swap handling**: When worktrees are at each other's expected locations (e.g., `alpha` @ `repo.beta`, `beta` @ `repo.alpha`), the command automatically resolves via a temp location.

**Main worktree**: When main worktree has a non-default branch checked out, creates a new linked worktree at the expected path and switches main back to default branch.

**Flags**:
- `--dry-run`: Preview what would be moved
- `--commit`: Auto-commit dirty worktrees before relocating  
- `--clobber`: Backup non-worktree paths to `<path>.bak-<timestamp>`

**Safety**: `--commit --clobber` should handle most scenarios without failing.

## Also includes

- Path redaction improvements for clearer snapshot output (`_PARENT_/<dir>` instead of `[PROJECT_ID]`)
- macOS path canonicalization fixes for `/var` → `/private/var` symlinks

## Test plan

- [x] 11 integration tests covering: no mismatches, single mismatch, dry run, locked worktree, target exists, clobber backup, dirty without commit, main worktree, swap scenario, specific branches, multiple worktrees
- [x] All 923 tests pass
- [x] Lints pass


max-sixty:
@claude move the plan to a spec in a rust module; then remove it

PR #802, opened 2026-01-22
feat: add shorthand aliases for list and remove commands

### LINKED ISSUE #406: Support shorthand
`wt ls` `wt rm` etc

zerone0x:
## Summary

Adds visible aliases `ls` for `list` and `rm` for `remove` commands, addressing the request in #406.

These are common shortcuts that users intuitively try after installing worktrunk (as noted in the issue comments). The aliases are shown in help text, making them discoverable:

```
Commands:
  switch  Switch to a worktree
  list    List worktrees and their status [aliases: ls]
  remove  Remove worktree; delete branch if merged [aliases: rm]
  ...
```

## Changes

- Added `visible_alias = "ls"` to the `List` command
- Added `visible_alias = "rm"` to the `Remove` command
- Updated snapshots for help text changes

## Notes

Using `visible_alias` (vs hidden `alias`) shows the aliases in help text, which helps discoverability while still allowing tab completion to work with the full command names. This addresses the maintainer's concern about tab completion from the issue discussion.

Fixes #406

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -410,6 +410,7 @@ To change which branch a worktree is on, use `git switch` inside that worktree.
 
     /// List worktrees and their status
     #[command(
+        visible_alias = "ls",
         after_long_help = r#"Shows uncommitted changes, divergence from the default branch and remote, and optional CI status.
 
 <!-- demo: wt-list.gif 1600x900 -->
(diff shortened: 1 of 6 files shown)


max-sixty:
thank you for the PR, but I don't think we're ready to make a decision here, I still think we should let people figure out their own setups for a while before potentially coalescing on an option.

Issue #801, opened 2026-01-22
Flexible commit-generation without "llm" or any external tool


can-sahin-gowago:
Cool tool, loving it. But I don't wanna install 'llm' just for generation commit messages. Already having that streamlined with skills etc with claude.  Why is something like this not possible (without installing any external tool)?

```toml
[commit-generation]
command = "claude"
args = ["-p"]
```

It gives me: `Unknown key commit-generation will be ignored`. In short, maybe you could create a flexibly configurable way of commiting?

max-sixty:
this is likely using the wrong config file; this should be in the user config file at `~/.config/worktrunk/config.toml`. I'm adding something that warns about this

(if that's not the reason, lmk...)

would be interested to see how you get on with `claude -p`. it's a great idea. my main concern is that it'll be quite slow. but maybe if we tell it to use the haiku model, disable thinking, etc; then it'll be quite reasonable...

max-sixty:
ok this was actually a very good idea indeed, and https://github.com/max-sixty/worktrunk/pull/809 moves us recommending this by default

it requires a slight change in how we do the LLM commit generation, so there's a deprecation of the confirm format

it's a bigger change so I'll leave it open for at least a day or so

can-sahin-gowago:
ok I see thanks. Im still not sure why UserConfig and Project config differs? I though they have exact same schema just User Config is overriding values from Project Config.

max-sixty:
user vs project are quite different, check out https://worktrunk.dev/config/#configuration-files, will also make this clearer

you're right we did recently add the ability to override project setting in the user config, in per-project sections. this is possibly confusing atm, I'll look to see if we can make it clearer

commit generation is a great example of why the configs are split: generally you want to define this for yourself rather than your team

lmk if that makes sense, thanks

can-sahin-gowago:
That is clear now but it was hard to get it the logic of config files. I think full example config file with all options explained with commands would make it simpler.

PR #804, opened 2026-01-22
Improve config validation warnings with cross-config suggestions

max-sixty:
## Summary

- Detect when config keys are placed in the wrong config file (user vs project)
- Suggest the correct location in warning messages
- Refactor to use generic traits instead of enum dispatch

**Example**: If `commit-generation` is found in project config, the warning now says:
> Key `commit-generation` belongs in user config (~/.config/worktrunk/config.toml) (will be ignored)

Instead of just:
> Unknown key `commit-generation` will be ignored

## Test plan

- [x] Unit tests for `warn_unknown_keys` with cross-config suggestions
- [x] Integration tests for `wt config` output with misplaced keys
- [x] All existing tests pass


max-sixty:
helps with #801

PR #814, opened 2026-01-23
Add wt sync command

davidbeesley:
## Summary

- Add `wt sync` as a first-class command (equivalent to `merge --no-remove`)
- Extract shared `MergeArgs` to reduce duplication

## Rationale

Some users prefer to sync their feature branches frequently throughout development rather than merging only at the end. This promotes that workflow to a dedicated command and first-class feature, increasing discoverability and ergonomics.

Acknowledging this is a UX trade-off, no worries if you prefer to keep it lean.

## Test plan

- [x] Help text snapshots for `sync -h` and `sync --help`
- [x] Functional test verifying sync behaves like `merge --no-remove`
- [x] Documentation auto-synced

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -308,7 +308,7 @@ impl MergeArgs {
     }
 }
 
-/// Convert paired flags (--flag / --no-flag) to Option<bool>
+/// Convert paired flags (--flag / --no-flag) to `Option<bool>`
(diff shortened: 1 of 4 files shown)


max-sixty:
thank you for the effort and sorry to be a downer, but this can just be an alias for users who want the alias; there is no need to have a top level command

PR #876, opened 2026-01-26
Fix worktree-path on tips page

uriahcarpenter:
Update tips page with correct `worktree-path` for bare repos. (Without `../` working dir is created inside `.git` directory.)

Second doc location missed in https://github.com/max-sixty/worktrunk/pull/320


max-sixty:
thanks a lot @uriahcarpenter !

Issue #924, opened 2026-02-03
`wt switch` interactive picker's diff preview layout incorrectly when using `delta`


tnlanh:
`wt switch` interactive picker's diff preview layout incorrectly when using `delta`. For the built-in git diff, it works normally.

Ref: https://github.com/dandavison/delta?tab=readme-ov-file#get-started

- Installation method: Homebrew
- Version: wt 0.22.0
- OS: macOS Tahoe 26.2
- Terminal emulator: Ghostty, iTerm2

<img width="1993" height="897" alt="Image" src="https://github.com/user-attachments/assets/1d6bc802-41f7-4ac8-98ca-6072ed5bd802" />

max-sixty:
sorry for not understanding — can you explain what is wrong with the diff? is it that it's side-by-side?

tnlanh:
I’ve configured Git to display the diff side by side using the `delta` option. However, the diff display in the `wt switch` picker doesn’t fill up the available space.

The diff in `git diff`:

<img width="1140" height="1402" alt="Image" src="https://github.com/user-attachments/assets/9160aa20-1fe9-4bb6-a9b4-8e21c7a1096b" />

The diff in `wt switch` picker:

<img width="1140" height="1402" alt="Image" src="https://github.com/user-attachments/assets/3bad51fc-4ba3-4fda-a4d3-dc15af10248a" />

tnlanh:
Hi Max, I’m trying to debug the issue using Claude and identify its root cause and solution. I’ll create a pull request for it.

```
The Fix
1. The run_git_diff_with_pager() function needs to pass the width to the pager. The fix would be to:
2. Accept a width parameter in run_git_diff_with_pager()
3. Set the COLUMNS environment variable when spawning the pager subprocess
4. Pass --width flag for pagers that support it (delta and bat both support --width)
```

<img width="1140" height="1402" alt="Image" src="https://github.com/user-attachments/assets/f5317712-2fed-4c55-a9c5-35572611de41" />

tnlanh:
PR: https://github.com/max-sixty/worktrunk/pull/925

PR #925, opened 2026-02-03
fix(pager): enhance `run_git_diff_with_pager` to accept width parameter

tnlanh:
This PR aims to resolve the issue with the diff pager using delta: https://github.com/max-sixty/worktrunk/issues/924: https://github.com/max-sixty/worktrunk/issues/924

- Updated `run_git_diff_with_pager` to include a `width` parameter, allowing for better control of output formatting.
- Adjusted the command pipeline to set the `COLUMNS` environment variable, ensuring pagers can detect the correct width for display.
- Improved documentation to clarify the behavior of the pager when no explicit configuration is provided.

This change aims to prevent hangs and ensure proper output formatting when using delta/bat/batcat pagers.

### DIFF
--- a/src/commands/select/pager.rs
+++ b/src/commands/select/pager.rs
@@ -238,19 +238,19 @@ mod tests {
     fn test_build_pager_command_with_delta() {
         // Note: This test assumes no explicit [select] pager config is set
         // If user has config, the behavior changes (returns pager_cmd as-is)
-        
+
         // delta should get --paging=never and --width
         let result = build_pager_command("delta", 120);
         if !has_explicit_pager_config() {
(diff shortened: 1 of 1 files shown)


max-sixty:
Superseded by #930 which takes a simpler approach — setting COLUMNS env var instead of auto-injecting --width flags. Thanks for the report and initial PR!

(from Claude)

max-sixty:
@tnlanh I think this is actually not possible to do in a general way (though ofc open if you have ideas...). Instead, the best option seems to be to just forward `$COLUMNS` and the consumer passes that into `--width`, or whatever their pager is. So #930 does that

lmk if that sounds reasonable, ofc always feel free to try to improve it...

PR #945, opened 2026-02-07
fix(list): filter bare repo entry from worktree list to prevent status errors

guillaumedeslandes:
Some git versions/configurations don't set the `bare` flag in porcelain output for certain bare repo patterns (e.g., project/.git as bare). This caused `wt list` to run `git status` on the bare repo directory itself, failing with "must be run in a work tree".

Add a secondary path-based filter in `list_worktrees()` that compares worktree paths against `git_common_dir` when the repo is bare, handling both direct match (path IS the bare dir) and parent match (path is parent of bare .git dir).

Original issue text proposed by Claude:

```
---
  Title: wt list runs git status on bare repo entry, causing "must be run in a work tree" error

  Body:

  Summary

  When using a bare repo setup, wt list attempts to run git --no-optional-locks status --porcelain on the bare repository itself (which appears in git worktree list output). This fails with "fatal: this operation must be run in a work tree".

  Reproduction

  1. Set up a bare repo with worktrees:
  git clone --bare <repo> project/.bare
  cd project/.bare
  git worktree add ../main main
  git worktree add ../feature feature-branch

  2. Run wt list from any worktree
  3. Observe error:
  ▲ Some git operations failed:
    main: working-tree-diff (fatal: this operation must be run in a work tree)

  Analysis

  From -vv output, wt correctly detects the bare repo:
  [e] $ git rev-parse --is-bare-repository [.bare]
  [e]   true

  But still runs status on it:
  [g] $ git --no-optional-locks status --porcelain [.bare]
  [g]   ! fatal: this operation must be run in a work tree

  The error reports as "main: working-tree-diff" because the bare repo's HEAD points to refs/heads/main.

  Expected behavior

  wt list should skip status/diff checks for bare repository entries detected via git rev-parse --is-bare-repository.

  Environment

  - wt version: (run wt --version)
  - git version 2.32.0
  - macOS

  ---
```

### DIFF
--- a/src/commands/config/show.rs
+++ b/src/commands/config/show.rs
@@ -490,8 +490,8 @@ fn render_project_config(out: &mut String) -> anyhow::Result<()> {
     }
 
     // Check for deprecations with show_brief_warning=false (silent mode)
-    // Only show in main worktree (where .git is a directory)
-    let is_main_worktree = repo_root.join(".git").is_dir();
+    // Only write migration file in main worktree, not linked worktrees
+    let is_main_worktree = !repo.current_worktree().is_linked().unwrap_or(true);
     let has_deprecations = if let Ok(Some(info)) = worktrunk::config::check_and_migrate(
         &config_path,
         &contents,
--- a/src/config/project.rs
+++ b/src/config/project.rs
@@ -134,10 +134,9 @@ impl ProjectConfig {
             .map_err(|e| ConfigError::Message(format!("Failed to read config file: {}", e)))?;
 
         // Check for deprecated template variables and create migration file if needed
-        // Only write migration file in main worktree (where .git is a directory)
-        // Linked worktrees have .git as a file pointing to the main worktree
+        // Only write migration file in main worktree, not linked worktrees
         // Use show_brief_warning=true to emit a brief pointer to `wt config show`
-        let is_main_worktree = repo_root.join(".git").is_dir();
+        let is_main_worktree = !repo.current_worktree().is_linked().unwrap_or(true);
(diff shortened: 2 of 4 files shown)


max-sixty:
> Some git versions/configurations don't set the `bare` flag in porcelain output for certain bare repo patterns (e.g., project/.git as bare).

Can I ask what these are? 

The change looks fine (and thank you for the change), but keen to ensure we only add things that are necessary, and I couldn't immediately see why git wouldn't supply the `bare` flag. 

ty!

max-sixty:
will close as stale but please feel free to reopen

guillaumedeslandes:
> > Some git versions/configurations don't set the `bare` flag in porcelain output for certain bare repo patterns (e.g., project/.git as bare).
> 
> Can I ask what these are?
> 
> The change looks fine (and thank you for the change), but keen to ensure we only add things that are necessary, and I couldn't immediately see why git wouldn't supply the `bare` flag.

Sorry for the delay,

The issue came from a repo I had Claude setup for a small side project. I wanted to work with workspaces, but move all git files out of the way, hence the `.bare` directory.
Setup reproduction is:
```
git clone --bare <repo> project/.bare
cd project/.bare
git worktree add ../main main
git worktree add ../feature feature-branch
```

Don't know if there are specific commands / outputs that would help you.
I'll try to redo this setup with a next project and reopen the PR if I have more exemples.

max-sixty:
thanks @guillaumedeslandes 

will push a test for this repro; if that closes this PR then feel free to reopen with another one. maybe it's a different version of git? or an older version of wt?

max-sixty:
Superseded by #1112 which adds the regression test from this PR.

PR #952, opened 2026-02-08
docs(hooks): add -sTCP:LISTEN to lsof in hook examples

andoniaf:
After being surprised a couple of times by a random browser tab closing when removing a worktree, I tracked it down to the `lsof` invocation in the hook examples.

   Without `-sTCP:LISTEN`, `lsof -ti :PORT` matches **any** process with a connection to that port — including browsers with an open tab to `localhost:PORT`. So the `xargs kill` was taking out the browser process along with the dev server.

   Adding `-sTCP:LISTEN` restricts the match to processes that are actually **listening** on the port (i.e., the dev server itself), leaving clients alone.
   
   I can't think of a case where using `-sTCP:LISTEN` would be worse than the current behavior, but I might be missing some edge case

   ## Changes

   - Added `-sTCP:LISTEN` flag to all `lsof` invocations in hook examples and project config"

### DIFF
--- a/.claude-plugin/skills/worktrunk/reference/hook.md
+++ b/.claude-plugin/skills/worktrunk/reference/hook.md
@@ -94,7 +94,7 @@ Cleanup tasks after worktree removal: stopping dev servers, removing containers,
 
 ```toml
 [post-remove]
-kill-server = "lsof -ti :{{ branch | hash_port }} | xargs kill 2>/dev/null || true"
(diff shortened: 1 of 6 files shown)


max-sixty:
great find, thanks @andoniaf !

andoniaf:
thanks to you @max-sixty, I love the project ❤️

PR #963, opened 2026-02-10
feat: add system-wide config file support

goodtune:
## Summary

- Add system-wide config file support for organization-wide defaults
- Config loading order: system config → user config → environment variables
- System config search locations follow platform conventions

## System config locations

- `$WORKTRUNK_SYSTEM_CONFIG_PATH` (explicit override)
- `$XDG_CONFIG_DIRS` directories (colon-separated, Unix)
- Linux: `/etc/xdg/worktrunk/config.toml`
- macOS: `/Library/Application Support/worktrunk/config.toml`
- Windows: `%PROGRAMDATA%\worktrunk\config.toml`

## Merge behavior

Uses the same format as user config. The config crate deep-merges TOML tables:
- **Different named hooks** from system + user coexist (e.g., org defines `company-lint`, user adds `my-lint`)
- **Same-named hooks**: user replaces system
- **Different hook types** are independent

Visible in `wt config show` with a hint when no system config exists.

## Prior art

- https://git-scm.com/docs/git-config#Documentation/git-config.txt---system
- https://code.claude.com/docs/en/settings#when-to-use-each-scope

## Test plan

- [x] Unit tests for config path resolution (XDG_CONFIG_DIRS, platform defaults, env override)
- [x] Integration tests for `config show` display (with/without system config, empty, invalid)
- [x] Integration tests for config value precedence (system defaults, user overrides)
- [x] Integration tests for hooks merge semantics (deep merge, same-name replacement, preservation)
- [x] Snapshot tests updated for all platforms (Linux, macOS, Windows)
- [x] All 1050 tests pass locally

### DIFF
--- a/src/config/user/path.rs
+++ b/src/config/user/path.rs
@@ -72,8 +72,7 @@ pub fn get_system_config_path() -> Option<PathBuf> {
     // Priority 2+3: Check XDG_CONFIG_DIRS (if set), otherwise platform defaults.
     // When XDG_CONFIG_DIRS is set, system_config_dirs() returns only those dirs
     // (per XDG spec, no fallback to platform defaults).
-    let (dirs, _) = system_config_dirs();
-    for dir in &dirs {
+    for dir in &system_config_dirs() {
         let path = dir.join("worktrunk").join("config.toml");
         if path.exists() {
             return Some(path);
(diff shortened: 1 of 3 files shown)


max-sixty:
thanks for the reformed PR

is this actually something that people want, though? I agree re the `git` comparison. But I don't think I've seen this much in other tools.

Could folks share some use cases where this would be helpful?

goodtune:
Here are some examples where it would be useful in our company.

We already deploy standardised configuration to all our developer machines to configure a proxy `ANTHROPIC_BASE_URL`, some OpenTelemetry config, etc in our `/etc/claude-code/managed-config.json`.

We have begun installing `worktrunk` on all their machines. We want to encourage them to use it to manage their agents worktrees, but also their interactive navigation. I have a couple of immediate use cases that I would like to standardise for them.

1. most of our projects need to collect dependencies from either submodules or some other package system based off a file in the repository
     - rather than having to ask every developer to configure `post-create` trigger to run `git submodule update --init` or `our-internal-tool-that-fetches-dependencies` we can do this in one place.
2. we have a hierarchy we are trying to encourage to help users get more familiar with worktrees
    - setting `worktree-path = "../{{ branch | sanitize }}"` in `/etc/xdg/worktrunk/config.toml` would allow us to not have to instruct every developer to do this themselves in `~/.config/worktrunk/config.toml`

Unless we have something like this, we need to ask users to copy our preferred configuration into their home directory. They don't always follow instructions to the letter, so when they try to follow along with internal guides they get confused because they missed a step.

The philosophy I think is very sound, and is echoed through specifications such as the XDG base directory standard, and tools such as [pip](https://pip.pypa.io/en/stable/topics/configuration/) (`/etc/pip.conf`), [uv](https://docs.astral.sh/uv/configuration/files/) (`/etc/uv/uv.toml`), [npm](https://docs.npmjs.com/cli/v10/configuring-npm/npmrc) (`/etc/npmrc`), [Cargo](https://doc.rust-lang.org/cargo/reference/config.html) (`/.cargo/config.toml`), [Docker](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file) (`/etc/docker/daemon.json`), [systemd](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html) (`/etc/systemd/`), [SSH](https://man.openbsd.org/ssh_config) (`/etc/ssh/ssh_config`), [Vim](https://vimhelp.org/starting.txt.html#system-vimrc) (`/etc/vimrc`), and [Ansible](https://www.redhat.com/en/blog/configuring-ansible) (`/etc/ansible/ansible.cfg`). This pattern exists precisely because it solves real operational problems in managed environments.

The ability to set sensible defaults system-wide while still allowing users to override them in their personal config is well-established and makes tools manageable in enterprise/team settings. Users who don't need this feature wouldn't be affected, but for organizations deploying tools at scale, it's the difference between a smooth rollout and constant support tickets.

max-sixty:
ok, that makes a lot of sense, thanks for the thoughtful response @goodtune 

let me make some small adjustments to the code, if that's OK

goodtune:
Absolutely, please tweak as needed!

goodtune:
Looks good, thanks for taking the time to improve it.

max-sixty:
hi @goodtune — wdyt about those changes? lmk if you think we can make any final improvements. ty!

goodtune:
Thanks @max-sixty looks good to me! Thanks for the revisions.

Issue #968, opened 2026-02-10
`wt switch` resolves symlinks, moving user from symlink path to canonical path


brooke-hamilton:
## Summary

When a user's working directory is accessed through a symlink, `wt switch` switches the user to the symlink-resolved path instead of the symlink path, resulting in the user's shell ending up at the resolved physical path instead of the equivalent path under the symlink.

## Reproduction

```bash
# Setup: symlink that maps a shorter path to a longer one
# /workspace/project -> /mnt/wsl/workspace/project  (symlink)

cd /workspace/project   # user is working through the symlink
wt switch feature

pwd
# Expected: /workspace/project.feature  (or equivalent under /workspace/)
# Actual:   /mnt/wsl/workspace/project.feature  (canonical path)
```

> Note: the above example represents my Windows/WSL setup in which I have a WSL Ubuntu instance running which I access from Windows Terminal. The `/mnt/wsl/workspace` --> `/workspace` is a symlink pointing to a VHD file that I mount onto multiple WSL instances. (In case you were wondering why I would want a symlink.)

## Root cause

Multiple places in the switch pipeline canonicalize paths, and the result is written directly to the directive file:

1. **Existing worktree paths** — `check_existing_worktree()` calls `dunce::canonicalize()` on the path from `git worktree list`
2. **New worktree paths** — `compute_worktree_path()` derives paths from `repo.repo_path()`, which uses `git rev-parse --show-toplevel` (returns canonical paths)
3. **Source worktree root** — `WorkingTree::root()` explicitly calls `canonicalize()`
4. **cwd detection** — `std::env::current_dir()` (which calls `getcwd()`) returns canonical paths on Linux

The user's **logical** path (preserving symlinks) is only available in the `$PWD` environment variable, which the binary never reads.

The directive file ends up containing:

```bash
cd '/mnt/wsl/workspace/project.feature'
```

instead of:

```bash
cd '/workspace/project.feature'
```

## Impact

This affects any user whose workflow relies on symlinks to access their projects — a common pattern with WSL, NFS mounts, and custom workspace layouts. After every `wt switch`, the user is silently moved out of their symlink tree, which can break relative paths, editor integrations, and muscle memory.

The same issue applies to `wt remove` and `wt merge`, which also write `cd` directives with canonical paths.

## Possible fix

[...]

PR #985, opened 2026-02-11
fix: use symlink-preserved paths in display messages

max-sixty:
## Summary

- PR #976 fixed `cd` directives to preserve symlink paths, but status messages still showed canonical paths (e.g., `✓ Created branch feature1 from main and worktree @ /mnt/wsl/workspace/...` instead of `@ /workspace/...`)
- Extracts `to_logical_path()` from `change_directory()` into a reusable function, applies it in `handle_switch_output()` and `print_switch_message_if_changed()` so display messages match the `cd` directive
- Adds positive and negative test assertions on stderr to verify display messages use the logical path

Ref: https://github.com/max-sixty/worktrunk/issues/968#issuecomment-3881595335

## Test plan

- [x] Existing symlink integration tests pass with new stderr assertions (both positive: contains symlink path, and negative: doesn't contain canonical path)
- [x] Full test suite passes (2327 tests, 0 failures)
- [x] All lints pass

### DIFF
--- a/src/output/global.rs
+++ b/src/output/global.rs
@@ -137,6 +137,22 @@ impl SymlinkMapping {
     }
 }
 
+/// Translate a canonical path to the user's logical (symlink-preserved) path.
+///
+/// If the user navigated via symlink (e.g., `/workspace/project` -> `/mnt/wsl/workspace/project`),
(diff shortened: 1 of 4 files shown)


max-sixty:
@brooke-hamilton lmk any gaps after this!

PR #1014, opened 2026-02-14
fix(plugin): remove duplicate skills declaration from marketplace.json

### LINKED ISSUE #1013: Plugin has conflicting manifests: both plugin.json and marketplace entry specify components
After installing via:
```
claude plugin marketplace add max-sixty/worktrunk
claude plugin install worktrunk@worktrunk
```

Claude Code warns:
> Plugin worktrunk has conflicting manifests: both plugin.json and marketplace entry specify components. Set strict: true in marketplace entry or remove component specs from one location.

`marketplace.json` declares `skills` with `"strict": false`, and `plugin.json` also declares `skills` and `hooks`. Both are specifying the same components, triggering the conflict.

**Fix**

Either:
- Set `"strict": true` in `marketplace.json` (lets `plugin.json` be the source of truth)
- Remove `"skills"` from `marketplace.json` since `plugin.json` already declares them

**Workaround**

```bash
sed -i '' 's/"strict": false/"strict": true/' ~/.claude/plugins/marketplaces/worktrunk/.claude-plugin/marketplace.json
```

**Environment**
- macOS
- Claude Code (latest)

jacksonblankenship:
- Removes `skills` and `strict` keys from the marketplace plugin entry, letting `plugin.json` be the single source of truth for component declarations
- Fixes Claude Code warning about conflicting manifests when installing via `claude plugin marketplace add`

Closes #1013


max-sixty:
thank you!

PR #1024, opened 2026-02-15
fix(statusline): scope branch diff to sparse checkout cone

bendrucker:
In sparse checkout repos, `branch_diff_stats()` runs `git diff --numstat merge_base..HEAD`, a commit-to-commit diff covering all files in the repo, not just the sparse cone. This produces inflated line counts (e.g. `^+4M -2M`) that don't reflect the user's working set.

## Changes

- Add cached `sparse_checkout_paths()` on `Repository` that runs `git sparse-checkout list` and returns cone paths (empty for non-sparse repos)
- Pass sparse checkout paths as pathspecs (`-- path1 path2`) to `git diff --numstat` in `branch_diff_stats()`
- Distinguish expected exit code 128 (non-sparse) from unexpected failures with `log::debug!`

Working tree diff (`@+N -N`) is already correct,`git diff --numstat HEAD` respects sparse checkout. Commit counts are unchanged since they count commits, not files.

## Testing

- `test_sparse_checkout_paths_empty_for_normal_repo`
- `test_sparse_checkout_paths_returns_cone_paths`
- `test_sparse_checkout_paths_cached`
- `test_branch_diff_stats_scoped_to_sparse_checkout`,verifies scoped stats are smaller than full-repo stats

### DIFF
--- a/src/git/repository/mod.rs
+++ b/src/git/repository/mod.rs
@@ -438,36 +438,17 @@ impl Repository {
     /// scoped to the worktree the user is running from, not per-listed-worktree.
     pub fn sparse_checkout_paths(&self) -> &[String] {
         self.cache.sparse_checkout_paths.get_or_init(|| {
-            let output = Cmd::new("git")
-                .args(["sparse-checkout", "list"])
-                .current_dir(&self.discovery_path)
-                .context(self.logging_context())
-                .run();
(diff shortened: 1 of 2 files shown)


max-sixty:
thank you!

I made a couple changes — let me know any feedback, otherwise will merge

bendrucker:
Thanks! My rust experience is minimal, this is definitely much better.

PR #964, opened 2026-02-10
Add nushell support

arnaudlimbourg:
This is my second attempt at having nushell support, referring #735 

This PR adds:
- nushell support
- tests for said support
- documentation updates

I tested it manually using nushell v0.110.0 and have not encountered issues.

PR #1032, opened 2026-02-15
fix(shell): auto-detect nushell for install even without vendor/autoload

arnaudlimbourg:
When `nu` is in PATH, `wt config shell install` now creates the vendor/autoload directory automatically instead of skipping nushell. The directory was introduced in nushell v0.96.0 and isn't always pre-created, causing install to silently skip nushell users.

Follows the existing PowerShell pattern: detect the shell binary, then set should_configure and allow_create so the directory gets created on the user's behalf.

This fixes issue reported in #964 about installation being skipped on nushell


arnaudlimbourg:
@max-sixty tested locally, works well

max-sixty:
thank you!

(fwiw I was happy to take a pass on this, but didn't finish before I went to bed, thank you for taking it...)

Issue #1036, opened 2026-02-15
Issue getting nushell configuration to work


arnaudlimbourg:
there is a lingering issue where nushell configuration isn't properly setup.

---
The log is fine now (updated to the last release too), but when I `wt switch -c somename` I get this:
```
✓ Created branch instush-fonts from main and worktree @ ~/fatbot.instush-fonts
▲ Cannot change directory — shell integration not installed
```

and just for context:
```
fatbot ➜ : wt config shell install
○ Already configured shell extension & completions for bash @ ~/.bashrc
○ Already configured shell extension & completions for zsh @ ~/.zshrc
○ Already configured shell extension for fish @ ~/.config/fish/functions/wt.fish
○ Already configured completions for fish @ ~/.config/fish/completions/wt.fish
○ Already configured shell extension for nu @ ~/.config/nushell/vendor/autoload/wt.nu
✓ All shells already configured
```

_Originally posted by @omerxx in https://github.com/max-sixty/worktrunk/issues/964#issuecomment-3904516980_

PR #1038, opened 2026-02-15
fix(shell): nushell detection checks multiple config paths

arnaudlimbourg:
When shell integration is installed, nushell config might be in different locations depending on the user's environment:
- Installation queries `nu -c "echo $nu.default-config-dir"` → creates file at that path
- Runtime check should find it regardless of which path was used

Previously, is_shell_configured() only checked the single path returned by nushell_config_dir(). If the `nu` command failed during runtime (PATH issues, timeout, etc.) but succeeded during install, detection would fail.

Fix: Return multiple candidate paths for nushell checking:
1. The path from querying nu (if available)
2. XDG_CONFIG_HOME/nushell (if set)
3. ~/.config/nushell (XDG default)
4. ~/Library/Application Support/nushell (macOS)

This ensures is_shell_configured() finds the nushell integration file regardless of which path was used during installation, fixing the issue where users got "Cannot change directory — shell integration not installed" despite having successfully configured nushell.

Resolves the follow-up issue from PR #1032 where nushell detection failed at runtime even after successful installation.

Refs #1036

### DIFF
--- a/src/shell/paths.rs
+++ b/src/shell/paths.rs
@@ -18,6 +18,15 @@ pub fn home_dir_required() -> Result<PathBuf, std::io::Error> {
     })
 }
 
+/// Parse the stdout of `nu -c "echo $nu.default-config-dir"` into a path.
+///
+/// Returns `Some(path)` if stdout contains a non-empty trimmed path, `None` otherwise.
+fn parse_nu_config_output(stdout: &[u8]) -> Option<PathBuf> {
+    let path_str = std::str::from_utf8(stdout).ok()?;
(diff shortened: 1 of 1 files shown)


arnaudlimbourg:
@max-sixty the code coverage was failing, it should be fine now. The claude-review is new I see but it fails with
```bash
Warning: Actor has insufficient permissions: read
Error: Action failed with error: Actor does not have write permissions to the repository
Error: Process completed with exit code 1.
```

max-sixty:
merging! later we can clean up the code a bit

PR #1050, opened 2026-02-16
fix(shell): nushell uninstall cleans all candidate locations, dedup candidates

max-sixty:
## Summary

Follow-up fixes to #1038 where `config_paths()` was changed to return multiple Nushell candidates but not all consumers were updated:

- **Uninstall iterates all candidate paths** instead of only `paths.first()`, so config files installed at a different location (e.g., `nu` was in PATH during install but not during uninstall) get cleaned up
- **`has_config_location` checks any candidate's parent** directory for auto-detection, not just the first
- **Deduplicated `nushell_config_candidates()`** — removed redundant etcetera entry (always duplicated either `$XDG_CONFIG_HOME/nushell` or `~/.config/nushell`) and added `HashSet`-based dedup for remaining edge cases

## Test plan

- [x] New integration test `test_uninstall_nushell_cleans_all_candidate_locations` — installs at default location, copies to secondary candidate, uninstalls with different `XDG_CONFIG_HOME`, verifies both files removed
- [x] New unit test `test_nushell_config_candidates_no_duplicates` — verifies no duplicate paths
- [x] Updated `test_nushell_config_candidates_always_has_fallback` — platform-aware assertions
- [x] All existing nushell tests pass (install, uninstall, auto-detection)
- [x] Full test suite (1033 tests), lints, clippy all green

### DIFF
--- a/src/commands/configure_shell.rs
+++ b/src/commands/configure_shell.rs
@@ -955,6 +955,12 @@ pub fn handle_unconfigure_shell(
     scan_for_uninstall(shell_filter, false, cmd)
 }
 
+/// Remove a config file with a context-rich error message.
+fn remove_config_file(path: &std::path::Path) -> Result<(), String> {
+    fs::remove_file(path)
+        .map_err(|e| format!("Failed to remove {}: {e}", format_path_for_display(path)))
+}
+
 fn scan_for_uninstall(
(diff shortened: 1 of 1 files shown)


max-sixty:
CC @arnaudlimbourg

PR #1077, opened 2026-02-17
Support local config

arnaudlimbourg:
If I want local, project-specific settings I need to either use the global config or stash between rebases and merges.

This pull request implements support for `.config/wt.local.toml` which you can `.gitignore`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -388,3 +388,137 @@ When adding new cached methods, see `RepoCache` in `src/git/repository/mod.rs` f
 ## Releases
 
 Use the `release` skill for cutting releases. It handles version bumping, changelog generation, crates.io publishing, and GitHub releases.
+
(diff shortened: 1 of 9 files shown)


max-sixty:
is there a big gap between this and the project-overides in the user configs?

the configs are already a bit complicated; I'm very keen to keep them simple...

arnaudlimbourg:
@max-sixty I understand. I wanted to have hooks that are very dependent on the installation and they don't make much sense for other users.

Maybe limiting extra local hooks would be better with a guard around not reusing an existing key?

max-sixty:
but what do the *overrides* not provide?

max-sixty:
am on phone but @claude could you show a couple examples of what I mean by overrides?

arnaudlimbourg:
@max-sixty that'll teach me to not read the doc carefully...

PR #1146, opened 2026-02-21
feat: recover gracefully when current worktree directory has been deleted

### LINKED ISSUE #1109: [Feature Request] Recover gracefully when current worktree directory has been removed
## Problem

When a linked worktree is removed (i.e via `wt remove` or `wt merge`  from another terminal or an agent action) while a shell is still in that directory, all `wt` commands fail:

```
❯ wt switch
✗ fatal: Unable to read current working directory: No such file or directory
```

This happens because `Repository::current()` calls `git rev-parse --git-common-dir` from `.` (the CWD), and the OS can't resolve a deleted directory. The user has to manually `cd` somewhere else before `wt` works again.

This is admittedly a minor annoyance.

## Proposed behavior

When `wt` detects the CWD no longer exists:

1. Read `$PWD` (shells preserve the logical path in this variable even after the directory is deleted, while `getcwd()` fails)
2. Load the user's `worktree-path` template.
3. Reverse-match the template against `$PWD` to infer the primary worktree path
4. Print a helpful message: `Current directory was removed. Try: cd ~/git/orchestrator`

For `wt switch` specifically, run the command as if it was invoked from the primary worktree instead of failing.

## Approach: template-to-regex reverse matching

Convert the user's `worktree-path` template into a regex, match it against `$PWD`, and use filesystem validation to find the primary worktree.

```
recover(deleted_path, template):
    expand ~ in template
    convert template variables to regex capture groups
    match regex against deleted_path

    if template uses "{{ repo_path }}/../" (sibling layout):
        # /../ is normalized away by OS, so capture "parent/repo_branch" as one string
        # split repo_branch on the separator (., -, _ — extracted from template)
        # try each split point left-to-right, check parent/candidate/.git exists
        # first hit is the primary worktree

    if template uses "{{ repo_path }}/..." (subdirectory layout):
        # repo_path is captured directly — return it
```

**Graceful fallback** — returns `None` (generic error) for templates with Jinja2 control flow, value-transforming filters, or no `repo_path` variable.

max-sixty:
## Summary

- When a linked worktree is removed from another terminal while a shell is still cd'd into it, `wt switch`, `wt list`, and the interactive picker now auto-recover by finding the parent repository from `$PWD`
- All other commands show a helpful hint: `Current directory was removed. Try: wt switch ^`
- Pre-switch hooks are skipped during recovery since the source worktree is gone
- Fixed a latent phantom blank line from `WorktrunkError::AlreadyDisplayed` empty Display impl

Thanks to @davidbeesley for reporting in #1109

Closes #1109

## Test plan

- [x] All 1084 integration tests pass
- [x] All 552 unit tests pass (6 new tests for recovery module)
- [x] Pre-commit lints clean
- [x] Snapshot updated for `merge_post_merge_command_failure` (hint now appears after CWD removal)
- [ ] Manual test: create worktree, delete it from another terminal, verify `wt switch ^` recovers

### DIFF
--- a/src/git/recover.rs
+++ b/src/git/recover.rs
@@ -337,13 +337,10 @@ mod tests {
     fn test_current_or_recover_returns_repo_when_cwd_exists() {
         // In a test environment, CWD exists, so current_or_recover should succeed
         // via the normal Repository::current() path (not recovery).
-        // This test will fail if not run inside a git repo, which is expected in CI.
-        if Repository::current().is_ok() {
-            let (repo, recovered) = current_or_recover().unwrap();
-            assert!(!recovered);
-            // Sanity check: repo should have a valid path
-            assert!(repo.repo_path().exists());
-        }
+        // Tests run inside a git repo in CI, so Repository::current() succeeds.
+        let (repo, recovered) = current_or_recover().unwrap();
+        assert!(!recovered);
+        assert!(repo.repo_path().exists());
     }
 
     #[test]
@@ -392,11 +389,7 @@ mod tests {
         std::fs::remove_dir_all(&wt_path).unwrap();
 
         // recover_from_path should find the parent repo
-        let recovered = recover_from_path(&wt_path);
-        assert!(
-            recovered.is_some(),
-            "should recover repo from deleted worktree path"
-        );
+        assert!(recover_from_path(&wt_path).is_some());
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


max-sixty:
@worktrunk-bot could you make this change if it still applies?

> The recovery-from-deleted-CWD pattern is now copy-pasted in three places (handle_switch, handle_select, main.rs list handler). Consider extracting a helper:

also fix ci

max-sixty:
great, do the consolation

max-sixty:
@worktrunk-bot 
- try fixing CI
- open an issue on @worktrunk-bot freqyently saying "Will approve once CI finishes.", but then going dark, have the issue-triager look at some cases and see whether we need to change our skill. it should be trying to fix CI

PR #1221, opened 2026-02-27
fix: propagate error from `is_bare()` instead of panicking

bendrucker:
`is_bare()` panics when `git rev-parse --is-bare-repository` times out because it uses `.expect()` on the `Cmd::run()` result. This changes the return type to `Result<bool>` so the error propagates instead.

I don't know why `git` was slow in this case, but it was:

```
thread '<unnamed>' (84001037) panicked at src/git/repository/mod.rs:459:18:
git rev-parse failed on valid repo: Custom { kind: TimedOut, error: "command timed out" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Rayon: detected unexpected panic; aborting
```

## Changes

- `is_bare()` returns `Result<bool>` using `get_or_try_init` (doesn't cache errors, so transient timeouts retry on next call)
- `repo_path()` uses `unwrap_or(false)` since it returns `&Path` via `OnceCell<PathBuf>` with many callers. Cascading `Result` would be disproportionate, and the non-bare fallback handles normal repos correctly
- All other callers already return `Result`, so they just add `?`

### DIFF
--- a/src/commands/step_commands.rs
+++ b/src/commands/step_commands.rs
@@ -1073,7 +1073,7 @@ pub fn handle_promote(branch: Option<&str>) -> anyhow::Result<PromoteResult> {
 
     // For normal repos, worktrees[0] is the main worktree
     // For bare repos, there's no main worktree - we don't support promote there
-    if repo.is_bare() {
+    if repo.is_bare()? {
         anyhow::bail!("wt step promote is not supported in bare repositories");
     }
 
--- a/src/commands/worktree/resolve.rs
+++ b/src/commands/worktree/resolve.rs
@@ -82,7 +82,7 @@ pub fn compute_worktree_path(
 ) -> anyhow::Result<PathBuf> {
     let repo_root = repo.repo_path();
     let default_branch = repo.default_branch().unwrap_or_default();
-    let is_bare = repo.is_bare();
+    let is_bare = repo.is_bare()?;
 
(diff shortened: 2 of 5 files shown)


bendrucker:
The `unwrap_or(false)` in `repo_path()` means that for bare repos, if `is_bare()` fails (same mystery git timeout), we fall through to `parent()` which gives the wrong cached path. Ideas to address this without changing the signature:

```rust
// Submodules: --show-toplevel succeeds (git has explicit core.worktree config)
if let Ok(out) = Cmd::new("git")
    .args(["rev-parse", "--show-toplevel"])
    .current_dir(&self.git_common_dir)
    .context(path_to_logging_context(&self.git_common_dir))
    .run()
    && out.status.success()
{
    return PathBuf::from(String::from_utf8_lossy(&out.stdout).trim());
}

// --show-toplevel failed:
// 1. Bare repos (no working tree) → git_common_dir IS the repo
// 2. Normal repos from inside .git → parent is the repo
match self.is_bare() {
    Ok(true) | Err(_) => self.git_common_dir.clone(),
    Ok(false) => self.git_common_dir
        .parent()
        .expect("Git directory has no parent")
        .to_path_buf(),
}
```

Figured I'd ask before implementing (or not).

max-sixty:
thanks a lot @bendrucker ! nice spot

I will merge this

@worktrunk-bot could you draft the follow-up change for our review please?

PR #1281, opened 2026-03-05
Fix repo name resolving to .git for bare-style worktree setups

### LINKED ISSUE #1279: In a bare repo layout the template var `repo` resolves to .git
I am using the following (bare repository layout](https://worktrunk.dev/tips-patterns/#bare-repository-layout):
```
myproject/
├── .git/       # bare repository
├── main/       # default branch
├── feature/    # feature branch
└── bugfix/     # bugfix branch
```

The template vars show repo=".git" — worktrunk is deriving the repo name from the  `main_worktree` path which is `.git`.

IMHO, this is a worktrunk bug in case the repo is a bare git directory at `myproject/.git`, so the repo name becomes `.git` instead of `myproject`.

I would expect the repo name come from the remote URL or the parent directory, because all repos with the same layout will have the same `repo` name "git". This makes something like [subdomain routing with caddy](https://worktrunk.dev/tips-patterns/#subdomain-routing-with-caddy) kind of pointless.

seakayone:
## Summary

- Fixes `{{ repo }}` template variable resolving to `.git` instead of the project name when using bare-style worktree setups (e.g. `/project/.git` as the common dir with sibling worktrees)
- When `repo_path()` detects a bare repo whose path ends in `.git`, it now returns the parent directory instead

Fixes #1279

## Test plan

- [ ] Verify `wt -vv list` shows correct `repo=` value in a bare-style worktree setup
- [ ] Verify hook templates using `{{ repo }}` produce the correct project name
- [ ] Verify actual bare repos (not ending in `.git`) still work correctly


max-sixty:
this might be a bit purist — but I'm a bit skeptical of choosing different paths based on the _name_. What if someone names it `.bare`? then we pick a different path?

I previously had avoided trying to identify the parent because I'm not sure there is a reliable way of doing that. I agree `.git` is a _reasonable_ heuristic, but it also seems like it would be quite confusing when the heuristic fails. Instead, the docs point to using a different worktree default path: https://worktrunk.dev/tips-patterns/#bare-repository-layout

Is there a better way? 

(and a disclaimer that I haven't got onto using bare repos, so I might be misunderstand)

seakayone:
Unfortunately git does not itself have the concept of a repo name. 
`wt` uses the folder name as a heuristic for determining it's `repo`.

The problem arises when someone checks out the repo in a folder which is named quite generic, like `.bare`, `.git` as described in the [wt tipps](https://worktrunk.dev/tips-patterns/#bare-repository-layout). And this is most probable if one is using the bare approach where a parent folder contains the bare repo in a hidden folder. Theoratically you can checkout the bare repo in a properly named folder `myproject` which would then contain the git internals and the worktree folders, this case is not a problem.

I think assuming that when the repo was checked out in a hidden folder starting with `.` might be better heuristic than only matching `.git`. If the folder is hidden, then choose the parent folder name.

Another approach might be to also look at the remotes and use their url for crafting a repo name. IMHO this is even more tricky as remotes are optional, there might be more than one and parsing the url is a challenge for itself.

One more option could be to make this configurable in the project's `wt.toml`.

max-sixty:
should we be relying on `{{ repo }}` in this case? in the docs [linked above] we suggest, for bare repos:

```
Configure worktrunk to create worktrees as subdirectories:

# ~/.config/worktrunk/config.toml
worktree-path = "../{{ branch | sanitize }}"
```

I'd be -1 on adding heuristics for paths. So I think that leaves two options:
- find a better default worktree path, that works with both bare and normal repos
- do it through config (current solution)

any ideas for the former? I agree we also can't use the remote for this, unfortunately. Is there any git command that can resolve `repo` for both bare and normal repos?

@worktrunk-bot any ideas??

max-sixty:
> 3. Improve the default worktree-path to not need {{ repo }} — The current default is {{ repo_path }}/../{{ repo }}.{{ branch | sanitize }}. For bare repos this produces nonsense regardless of heuristics. A bare-repo-aware default (or just documenting the override, which the docs already do) sidesteps the problem.

right, but are there viable alternatives that would produce the same result for normal repos and reasonable results for bare repos?

max-sixty:
if it's driven by just "is it bare", then we could just have `repo` to mean the parent in the bare case

I thought that the underlying problem here was that there are _two common bare layouts_, is that not right?

could you lay out all the common layouts and how our variables map against them?

max-sixty:
ok, so, my current thinking:
- defining `repo` as `repo_path | parent_if_hidden | name` is possible...
  - or even `repo_path | parent_if_hidden_and_bare | name`
- but it's going to be _extremely_ confusing if anyone ever has a repo named `.foo`, which is not impossible (a dotfiles repo?)
- adding a `name` config to the project is not ideal, because these are all user configs, and then projects need a new config to fully support worktrunk, which is only useful in a tiny number of cases

is there any synthesis of these problems that solves it?!

seakayone:
@max-sixty I really appreciate your interest in solving this problem and completely understand that a heuristic taking "hidden" files into account does not work for you. 

The real problem is trying to have a repo name when git itself does not have the concept of such a name. So any heuristic is potentially inaccurate. Mostly people check out the repo in the same folder as in the url, that is why using the url for bare repos is a good candidate.

Using the remote url of origin would also work for me and the origin remote is the git standard when checking out a new repo so it is available in most cases and would also solve my problem.

max-sixty:
> coming back to this given the comments on https://github.com/max-sixty/worktrunk/issues/1279, even though I see it's been closed — not sure if that was deliberate?

thanks @seakayone, appreciate your understanding. I know my concerns are with edge cases (but one hundred 1% edge cases means something is getting hit a lot of the time...)

I'm _also_ skeptical of using the remote; in particular this often hits when _first_ using a repo, when a default remote may not have been set up; and then we get different behavior on something as fundamental as "where worktrees go" as soon as we set up a remote?

I think we may need to escalate this to the user; I can see a couple options:
- Print a warning when `repo` resolves to a hidden dir
- Offer a "seems like this is a bare repo, do you want to set the `worktree-path` to [.git/..] for this repo" (will think of the exact phrase)
- Any others?

Xilis:
Hey!

I'm not fully sure if this is considered finished, please let me know if I should instead continue in #1279 or if you are not considering further changes here.

Currently, using a bare repo setup like

```
project/
  .git
```

you get prompted if you want

```
worktree-path = "{{ repo_path }}/../{{ branch | sanitize }}"
```
which works if you want the folder structure of

```
project/
  .git
  something
  somethingelse
```

However, if you want

```
project/
  .git
  project.something
  project.somethingelse
```

(or something similar to that) I think that is currently not possible, since `{{ repo }}` resolves to `.git`.

Based on the discussion so far, would something like just exposing the parent folder (`project` in this case) as a template variable (`{{ parent_dir }}` or whatever) be a reasonable solution?

max-sixty:
great point @Xilis ! this should be fixed in https://github.com/max-sixty/worktrunk/pull/2592#pullrequestreview-4222628920

PR #1319, opened 2026-03-06
Use current_branch instead of target_branch in squash message

ricafeal:
- "merge" command can include a target branch, for example for situations where we are working in a branch that has been created from another branch
- In those cases, the "merge" command will create a merge commit that will include the target branch in the commit message
- We want to have the source branch in the merge commit message instead of the target branch


max-sixty:
thank you @ricafeal !

Issue #133, opened 2025-12-10
Windows: wt naming conflict


ctolkien:
`wt` already exists on Windows platforms for "Windows Terminal".

May need to consider adding an alias on these platforms.

Ref #121

ctolkien:
I would suggest just going with `worktrunk` and then users can add their own should they wish to.

max-sixty:
eeek, that is unfortunate. I did a scan for tools at first but didn't catch this.

let me reflect. I did like `wt`. I think having something standard that's short is an advantage vs. aliases, so examples look good / sharing commands can share a command that works, etc

max-sixty:
one option would be `git-wt` and then `git wt switch` works...

shanselman:
ya I just hit this as well, worktrunk (wt) is now installed itself ahead of windows terminal in the path :(. I'm aliasing worktree -> git-wt

max-sixty:
@shanselman & @ctolkien (& any other windows users)

what do you think would be a better approach, without abandoning `wt` totally?

how do most folks install things on Windows? should we have the winget install default to `git-wt`? add an alias like `wt_` / `wot`?

I admittedly hardly use Windows, but keen for Worktrunk to be helpful to Windows users, as well as good citizen in the ecosystem...

NikiforovAll:
Hi there, also impacts me. I think `worktrunk` is a good default name. 

 People can create aliases for better ergonomics depending on how they want to use `worktrunk`.  I ended up with https://github.com/NikiforovAll/dotfiles2/blob/main/dot_bashrc#L114-L124

ctolkien:
> [@shanselman](https://github.com/shanselman) & [@ctolkien](https://github.com/ctolkien) (& any other windows users)
> 
> what do you think would be a better approach, without abandoning `wt` totally?

Honestly, the "Windows" way  (or more specifically, the Powershell way, which I'd wager is what most Windows users are using for a shell), is to not abbreviate by default, but introduce an alias should you wish to.

max-sixty:
so this issue has the latest: `git-wt` is now installed by default in the WinGet package

ctolkien:
> so this issue has the latest: `git-wt` is now installed by default in the WinGet package

`git-wt` it is then?

shanselman:
yes, but also note in the Readme or docs that windows users can just turn off App Execution Aliases if they really want WT to be WorkTrunk

[...]

PR #1340, opened 2026-03-07
fix(plugin): use `git-wt` instead of `wt` in hook commands

sameersadiq1:
## Summary

- Replace `wt` with `git-wt` in all three Claude Code hook commands
- On Windows, `wt` resolves to `wt.exe` (Windows Terminal) instead of worktrunk, causing `error 2147942402 (0x80070002)` on every session exit
- The `wt` shell alias (from `wt config shell install`) isn't available in hooks since they spawn subprocesses, not interactive shells

Refs #133

## Why `git-wt` and not `wt`

`wt` resolves on a user's machine in two ways:

1. **Shell function** — injected into `.bashrc`/`.zshrc` by `wt config shell install`. Only available in interactive shells.
2. **Homebrew symlink** — `wt` → `git-wt` in `/opt/homebrew/bin/`. A convenience from one package manager on one platform.

Claude Code hooks use `"type": "command"`, which spawns a subprocess — no `.bashrc` sourced, no shell function. The only reason `wt` works on macOS today is the Homebrew symlink.

| Install method | Binary name | `wt` available? |
|---|---|---|
| Homebrew | `git-wt` | Yes (symlink) |
| Winget | `git-wt` | No (`wt.exe` = Windows Terminal) |
| Cargo install | `git-wt` | No |
| Manual | `git-wt` | Only if user creates alias |

`git-wt` works in 4/4 cases. `wt` works in 1/4. There is no scenario where a user has `wt` on PATH but not `git-wt` — the reverse is common.

## Test plan

- [x] `git-wt config state marker set/clear` works from a subprocess on Windows
- [ ] Install plugin locally, start and exit a Claude Code session — no Windows Terminal error
- [ ] `git-wt list` shows marker set/cleared properly

### DIFF
--- a/src/git/repository/branches.rs
+++ b/src/git/repository/branches.rs
@@ -24,12 +24,13 @@ impl Repository {
             .is_ok())
     }
 
-    /// Get all branch names (local branches only).
+    /// List all local branch names, sorted by most recent commit first.
     pub fn all_branches(&self) -> anyhow::Result<Vec<String>> {
         let stdout = self.run_command(&[
-            "branch",
+            "for-each-ref",
             "--sort=-committerdate",
             "--format=%(refname:lstrip=2)",
+            "refs/heads/",
         ])?;
         Ok(stdout
             .lines()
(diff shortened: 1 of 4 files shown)


max-sixty:
> **Homebrew symlink** — `wt` → `git-wt` in `/opt/homebrew/bin/`. A convenience from one package manager on one platform.

@worktrunk-bot is this accurate?

max-sixty:
@worktrunk-bot but not the central brew tap!

Closing this, as I don't think there's a way to do this universally, and it would be overly complex code to check for binaries in the hooks.

for those who want to keep `git-wt`, copying the hooks is viable

(if lots of people 👍 this issue and then someone pings me, I will change it!)

PR #1401, opened 2026-03-09
feat(switch): add no-cd config option

jradtilbrook:
## Summary
- Add `[switch] no-cd = true` config option to disable directory change by default
- Add `--cd` flag to override config when directory change is desired
- Implement 3-way precedence: `--cd` flag → `--no-cd` flag → config → default (false)

## Motivation
Users who manage worktrees via tmux sessions (one session per worktree) handle directory changes themselves. For these workflows, having `wt switch` automatically change directories is unnecessary — they switch tmux sessions instead. This config option lets users disable the default cd behavior while retaining `--cd` for occasional use.

### DIFF
--- a/src/commands/select/mod.rs
+++ b/src/commands/select/mod.rs
@@ -41,7 +41,7 @@ enum PickerAction {
 pub fn handle_select(
     cli_branches: bool,
     cli_remotes: bool,
-    change_dir: bool,
+    change_dir_flag: Option<bool>,
 ) -> anyhow::Result<()> {
     // Interactive picker requires a terminal for the TUI
     if !std::io::stdin().is_terminal() {
@@ -50,8 +50,9 @@ pub fn handle_select(
 
     let (repo, is_recovered) = current_or_recover()?;
 
-    // Merge CLI flags with resolved config
+    // Merge CLI flags with resolved config (project-specific config is now available)
     let config = repo.config();
+    let change_dir = change_dir_flag.unwrap_or_else(|| !config.switch.no_cd());
     let show_branches = cli_branches || config.list.branches();
     let show_remotes = cli_remotes || config.list.remotes();
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -462,7 +462,7 @@ fn handle_list_command(
 fn handle_select_command(branches: bool, remotes: bool) -> anyhow::Result<()> {
     // Deprecated: show warning and delegate to handle_select
     warn_select_deprecated();
-    handle_select(branches, remotes, true)
+    handle_select(branches, remotes, None)
 }
 
 #[cfg(not(unix))]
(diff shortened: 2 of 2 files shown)


docs/content/switch.md
```diff
@@ -211,11 +211,17 @@ Usage: <b><span class=c>wt switch</span></b> <span class=c>[OPTIONS]</span> <spa
       <b><span class=c>--clobber</span></b>
           Remove stale paths at target
 
+      <b><span class=c>--cd</span></b>
```
max-sixty: @worktrunk-bot can you hide this option and put it into `--no-cd`, in the same way we do with `--progressive` (i.e. keep `no-cd`

(@jradtilbrook lmk if you disagree, you're welcome to revert)

jradtilbrook: No i was actually going to ask that. I think its a better approach 👍

max-sixty: @worktrunk-bot ?

cperalt:
just for my own learning, I also use tmux sessions + worktrees, and i use have an alias like:

alias wts="wt switch --no-cd"
alias wtn="wt switch --create --no-cd"

to use when I dont want to switch directories. Does this not accomplish the same thing? Or is it that user wants no cd to be the default for his config?

jradtilbrook:
> just for my own learning, I also use tmux sessions + worktrees, and i use have an alias like:
> 
> alias wts="wt switch --no-cd" alias wtn="wt switch --create --no-cd"
> 
> to use when I dont want to switch directories. Does this not accomplish the same thing? Or is it that user wants no cd to be the default for his config?

@cperalt basically, I didn't want to add an alias 🤷‍♂️. I don't think there will ever be a time where I would want to allow it to cd either so wanted it globally configurable

cperalt:
True thanks for the reply, yeah this makes sense! Ill probably also use this too since i also never cd lol

docs/content/switch.md
max-sixty: @worktrunk-bot all good bud?

(For the hourly reviewer — please evaluate what happened here!)

max-sixty:
> just for my own learning, I also use tmux sessions + worktrees, and i use have an alias like:

ha, thanks, that makes for a more compelling reason to merge, if multiple people are setting up aliases!

docs/content/switch.md
max-sixty: @worktrunk-bot the issue re you not seeing messages on fork-pr inline comments should now be fixed, please now do this

max-sixty:
@worktrunk-bot please look at the messages above (I know you can't see inline review comments on forks...)

max-sixty:
@worktrunk-bot are the per-project overrides now working?

anything else to consider before we merge?

max-sixty:
> The only place it doesn't apply is the picker path (no branch argument), which resolves config with resolved(None) before a repo is opened — documented with a comment at main.rs:499-502. That's a reasonable limitation since the picker is repo-agnostic at that point.

is this pre-existing? open a new PR to address it if so

max-sixty:
can you push to this PR? I don't think the stacked PR is working given it's a fork

PR #1404, opened 2026-03-09
feat(switch): add --print flag for scripting

ruudk:
First of all, thanks for this great project! I just started with it a few days ago and I love it!! 

I'd like to be able to script a bit more and thought it would be nice to include a `--print` option to the `wt switch` command.

It opens the interactive picker and then when you press enter, it prints the selected working tree back. This can be used to automate/script things easier for custom workflows.

Curious what you think :)

### DIFF
--- a/src/commands/select/mod.rs
+++ b/src/commands/select/mod.rs
@@ -314,27 +314,12 @@ pub fn handle_select(
 
         // --print: just output the selected branch name and exit (read-only, no side effects)
         if print_only {
-            let identifier = match action {
-                PickerAction::Create => {
-                    let query = out.query.trim().to_string();
-                    if query.is_empty() {
-                        anyhow::bail!("Cannot create worktree: no branch name entered");
-                    }
-                    query
-                }
-                PickerAction::Switch => {
-                    let selected = out
-                        .selected_items
-                        .first()
-                        .context("skim accept has no selection")?;
-                    selected.output().to_string()
-                }
-                PickerAction::Remove => {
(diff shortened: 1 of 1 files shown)


ruudk:
@max-sixty It's finally ready 😁

max-sixty:
I was trying to think whether there was a way of getting this without adding another option. 

we could add a shortcut like `alt-p`. do you think that would suffice? or it's going to lead to lots of cursing when you inevitably hit 'return' and it goes and switches branch? 😀

possibly we could default to this when we run the picker with `--no-cd`?!

ruudk:
> we could add a shortcut like `alt-p`. do you think that would suffice? or it's going to lead to lots of cursing when you inevitably hit 'return' and it goes and switches branch? 😀

Hmm, if you run `wt switch --print` and press enter, it should not switch anything. It just prints the worktree name.

ruudk:
Another option could be, to add this to `wt list` and make that one interactive, so that it can be used for scripting?

max-sixty:
@worktrunk-bot can you draft a PR that, if we run `wt switch --no-cd`, then the interactive picker only prints, rather than changing path? (and then we can avoid adding the `--print` option)?

max-sixty:
I prefer https://github.com/max-sixty/worktrunk/pull/1445 — @ruudk are you OK with that? we get the same feature without another arg

ruudk:
Perfect let's do that one! Thank you 😊

PR #1438, opened 2026-03-11
merge: add --no-ff flag for merge commit (semi-linear history)

siriobalmelli:
## Summary

- Add `--no-ff` / `--ff` CLI flags to `wt merge`
- Add `[merge] no-ff` config option (project and user config)
- Refactor `handle_push` into a reusable `MergeContext` struct, eliminating duplication between fast-forward and merge-commit paths
- Create merge commits via git plumbing (`commit-tree` + `update-ref`) without checking out the target branch

## Motivation

I prefer rebased commits on the feature branch, plus a merge commit on the target.

This is equivalent to GitLab's "Merge commit with semi-linear history" strategy and GitHub's "Create a merge commit" option (after rebasing).

## Pipeline behaviour with `--no-ff`

The merge pipeline runs identically up through rebase and pre-merge hooks. At the merge step, instead of fast-forwarding the target ref, a merge commit is created:

```
git commit-tree HEAD^{tree} -p <target-tip> -p HEAD -m "Merge branch '<branch>' into <target>"
git update-ref refs/heads/<target> <new-merge-commit> <old-target-tip>
```

The target worktree (if present) is then synced via `reset --hard`.

## Config

```toml
[merge]
no-ff = true   # always create a merge commit
```

Override per-invocation with `--ff` to force fast-forward even when config says `no-ff = true`.

## Test coverage

11 integration tests covering:
- Basic single-commit `--no-ff`
- Multi-commit `--no-ff` (no squash)
- `--no-ff` with squash (squash first, then merge commit)
- `--no-ff` from config, with `--ff` CLI override
- `--no-ff` with rebase (diverged branches)
- `--no-ff` already up to date
- `--no-ff` with diverged branches and `--no-rebase` (error path)
- `--no-ff` syncs target worktree via `reset --hard`
- `--no-ff` with non-overlapping dirty target (autostash/restore)
- `--no-ff` with overlapping dirty target (safe failure, no data loss)
- `--no-ff` when target has no checked-out worktree

## Notes

I developed this PR with assistance from AI.

### DIFF
--- a/src/commands/worktree/push.rs
+++ b/src/commands/worktree/push.rs
@@ -10,7 +10,7 @@ use anyhow::Context;
 use color_print::cformat;
 use worktrunk::git::{GitError, Repository};
 use worktrunk::styling::{
-    eprintln, format_with_gutter, info_message, progress_message, success_message,
+    eprintln, format_with_gutter, info_message, progress_message, success_message, warning_message,
 };
 
 use super::types::MergeOperations;
@@ -365,10 +365,14 @@ pub fn handle_no_ff_merge(
     {
         let target_wt = ctx.repo.worktree_at(wt_path);
         if let Err(e) = target_wt.run_command(&["reset", "--hard", "HEAD"]) {
-            log::warn!(
-                "Failed to sync target worktree at {}: {e}. Run `git reset --hard HEAD` there manually.",
-                wt_path.display()
+            eprintln!(
+                "{}",
+                warning_message(cformat!(
+                    "Failed to sync target worktree; run <bold>git -C {} reset --hard HEAD</> manually",
+                    worktrunk::path::format_path_for_display(wt_path)
+                ))
             );
+            log::warn!("Failed to sync target worktree: {e}");
         }
     }
 


max-sixty:
how much do folks want this?

if there is a broad support then OK. but otherwise I would claim that the use case for `wt merge` — generally personal projects where we favor moving quickly & the transparency of commit histories (over complicated & precise commit histories) — means that squashing is almost always better

...and we want to avoid collecting all the git options in `wt`

...and when there's a special circumstance, the plain git commands are there...

siriobalmelli:
> how much do folks want this?
> 
> if there is a broad support then OK. but otherwise I would claim that the use case for `wt merge` — generally personal projects where we favor moving quickly & the transparency of commit histories (over complicated & precise commit histories) — means that squashing is almost always better
> 
> ...and we want to avoid collecting all the git options in `wt`
> 
> ...and when there's a special circumstance, the plain git commands are there...

Those are valid concerns.

I can't speak for other devs/teams, in my org this is unfortunately a non-negotiable workflow contract.

To me, the high appeal of `wt` is exactly the fact that it's a single, short command every time - and it just "does the right thing" ... the minute I need to start scripting my way around merge, I may as well go back to the morass of git aliases I was using previously.

The approach I'm proposing is that this is entirely opt-in: unless the user looks for `--no-ff`, it will not get in their way or otherwise complicate matters.

I'm happy to go the extra mile and make sure this is easy to merge and very maintainable, or otherwise if an issue comes up later I'll still be here and willing to find fixes. I'm also driving adoption into nixpkgs (https://github.com/NixOS/nixpkgs/pull/496607) and am volunteering to maintain it there too.

In the meantime, I've reworked the PR to fix test issues and improve testing, will to another run to see if something can be done about the slight drop in coverage.

max-sixty:
@siriobalmelli OK, I appreciate that, and also would like to reciprocate your contribution to the project...

one quick idea (not saying this would be better, but exploring in pursuit of avoiding more config options) — @worktrunk-bot can you assess the idea of a `git-options` config to `wt merge`, which in this case could include `--no-ff`? what else could be useful for folks to put in that? we could also allow this to be passed after `--` potentially

siriobalmelli:
> @siriobalmelli OK, I appreciate that, and also would like to reciprocate your contribution to the project...
> 
> one quick idea (not saying this would be better, but exploring in pursuit of avoiding more config options) — @worktrunk-bot can you assess the idea of a `git-options` config to `wt merge`, which in this case could include `--no-ff`? what else could be useful for folks to put in that? we could also allow this to be passed after `--` potentially

I appreciate that. Ironically, this was also my first instinct: let's just honor git options passed by the caller.

On closer inspection, my thinking is that the exact value add of `wt` (interface simplicity) stems in large part from the fact it doesn't precisely mirror a series of `git` commands (otherwise, it would be a trivial scripting problem, and it isn't) - and so it's not straightforward to just pass flags through.

BTW I'm running my branch on multiple machines and it's working great.

max-sixty:
OK!

I notice it's a big diff! Is this because we can't use the "internal" `push` approach? Could we describe the approach this takes?

siriobalmelli:
Rebased onto latest `main` (was 91 commits behind). All conflicts resolved — the core implementation (`push.rs`, `merge.rs`) applied cleanly since upstream hadn't touched those files; only `src/cli/mod.rs` and `src/main.rs` needed manual conflict resolution to incorporate upstream additions (`no-cd`, `step eval`, `RemovePlans`, etc.).

All 1668 tests pass locally (466 unit + 1202 integration). CI should be green once it runs.

Happy to address any remaining concerns — is there anything else you'd like changed before merging? @max-sixty

siriobalmelli:
Fixed in the latest commit (`17fb459b`): added a `warning_message` to stderr (always visible to the user) and demoted the `log::warn!` to debug-only. The message now reads:

> ⚠ Failed to sync target worktree; run **git -C \<path\> reset --hard HEAD** manually

Addresses the review at https://github.com/max-sixty/worktrunk/pull/1438#discussion_r2939581975.

siriobalmelli:
@max-sixty I'm confident this is ready to merge and will work fine in the long term.

All of my workflows internally have been using this for over a week with no trouble.

max-sixty:
thanks @siriobalmelli !

nirvdrum:
This has already been merged, so maybe you don't need any more feedback, but I have two primary use cases:

* open-source projects where I don't get to dictate the merge policy
* personal projects, which unfortunately often sit latent while life interferes, so the git history is very important context when I pick something back up (squashing complicates that a fair bit)

So, thank you for merging this. I was afraid my workflow was going to have to entail managing all of this outside of worktrunk. That workflow is scriptable, but I'm glad I don't have to do that.

PR #1454, opened 2026-03-11
fix: add --no-session-persistence to Claude commit generation

max-sixty:
Without this flag, `claude -p` saves the commit generation conversation to session history. When the user later runs `claude --continue`, it resumes that ephemeral commit session instead of their previous working session.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -385,7 +385,7 @@ tasks:
 
         # Documented commands from llm-commits.md
         declare -A COMMANDS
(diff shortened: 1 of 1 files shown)


max-sixty:
> The same Claude command in [`Taskfile.yaml`](https://github.com/max-sixty/worktrunk/blob/71f602118fea22aa45afa16f6e70573c4c02b02a/Taskfile.yaml#L388) (LLM benchmark task) is missing `--no-session-persistence`. Happy to push a fix commit if you'd like.

yes please

PR #1465, opened 2026-03-12
fix: suppress GPG signature output in all git log calls

apre:
# Problem

`wt list` fails on repositories where `log.showSignature` is set. When git logs a signed commit with `--format=%ct %s`, it emits GPG verification lines to stdout before the format line:

```
gpg: Signature made Wed Mar 11 18:18:28 2026 CET
gpg:                using EDDSA key 0000...
gpg: Good signature from "A U Thor <author@example.com>" [ultimate]
1773249428 fix: some commit message
```

The parsing code expected only the format output, so `split_once(' ')` would fail on the `gpg:` prefix.

# Fix

Pass `--no-show-signature` to all `git log` calls to suppress GPG verification output at the source. This protects all four call sites (`commit_details`, `commit_timestamps`, `commit_subjects`, `recent_commit_subjects`) rather than adding GPG-aware parsing to just one.

# disclaimer

The original fix, tests, and PR were fully generated by claude. The revision to use `--no-show-signature` was also applied by claude based on reviewer feedback.

### DIFF
--- a/src/commands/select/items.rs
+++ b/src/commands/select/items.rs
@@ -377,6 +377,7 @@ impl WorktreeSkimItem {
         let args = vec![
             "log",
             "--graph",
+            "--no-show-signature",
             format,
             "--color=always",
             "-n",
--- a/src/git/repository/diff.rs
+++ b/src/git/repository/diff.rs
@@ -85,8 +85,13 @@ impl Repository {
         // is the first line only (no embedded newlines). Split on first space.
         // --no-show-signature suppresses GPG verification output that otherwise
         // contaminates stdout when log.showSignature is set.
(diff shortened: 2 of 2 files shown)


max-sixty:
@worktrunk-bot should we be using something more porcelain-like, or disabling configs, in order to get a reliable output from this command?

max-sixty:
> Yes — `--no-show-signature` is the right root-cause fix here. I've pushed a revision that takes that approach instead.

but is this just one thing that might cause differences in log outputs? and we should have a command that is not dependent on local configs?

apre:
thank you @max-sixty  ! 🙏

PR #1506, opened 2026-03-14
docs(list): rewrite LLM summaries section

max-sixty:
Remove the TODO ("promote this feature more prominently once it's been tested in the wild") since LLM summaries are now on the homepage. Tighten the subsection — the prerequisites are already in the Columns table above, so the text now focuses on the mechanism (reuses `commit.generation`), how to enable (`summary = true`), and caching behavior.


max-sixty:
> the new wording drops the --full prerequisite

create a follow-up PR

PR #1541, opened 2026-03-15
Add switch --force-create option

sysradium:
I found myself systematically forgetting if I had already created a branch or not. Maybe it is just me, but that kept annoying me.

So I propose an ability either to allow a user to do it once via `--force-create` or override this behaviour globally via config.

It can be overcome with a tiny bash script/alias in zshrc. So this is just a suggestion.


max-sixty:
Can I suggest trying this with an alias for a while, if you find it great then come back and propose it as an option?

We had this discussion right at the beginning (@worktrunk-bot can probably find the issue). I'm keen to try to limit the growth of options, and this also has the higher bar given it is incongruous with git...

PR #1736, opened 2026-03-25
fix: suppress false positive shell integration warning for wrapper files

### LINKED ISSUE #1735: `wt config show` emits false positive "not detected as integration" warning for Nushell
`wt config show` correctly identifies the Nushell integration and prints:

```nushell
○ nu: Already configured shell extension & completions @ ~/.config/nushell/vendor/autoload/wt.nu
```

But then also emits a warning for the same file:

```nushell
▲ Found wt in ~/.config/nushell/vendor/autoload/wt.nu:7 but not detected as integration:
  export def "nu-complete wt" [context: string] {
  let external = (which -a wt | where type == "external")
  export def --env --wrapped wt [...args: string@"nu-complete wt"] {                                                                                                      let external = (which -a wt | where type == "external")
  msg: "wt binary not found in PATH. Install with 'cargo install --path .' or set $env.WORKTRUNK_BIN"
```

The flagged lines are source code from the integration file itself: variable lookups, string literals, and function definitions that happen to contain `wt` at a word boundary. The integration is working correctly; the warning is a false positive.

saschabratton:
**Problem**: `wt config show` emitted a false positive "Found wt in ... but not detected as integration" warning for Nushell (and Fish) wrapper files. These files ARE the integration (they contain the shell function definition), not config files that source it, so `is_shell_integration_line()` correctly doesn't match their contents, but the warning logic didn't account for this.

**Fix**: Before emitting unmatched candidate warnings, collect paths already confirmed as integration files by `scan_shell_configs`. For Fish/Nushell (standalone wrapper files), any action means the file was recognized. For eval-based shells (Bash/Zsh), only `AlreadyExists` counts. Skip those paths when checking for unmatched candidates.

Fixes #1735

### DIFF
--- a/src/commands/config/show.rs
+++ b/src/commands/config/show.rs
@@ -895,8 +895,7 @@ fn render_shell_status(out: &mut String) -> anyhow::Result<()> {
             // For wrapper-based shells, the file at the path IS the integration — any
             // action means it was recognized. For eval-based shells, only AlreadyExists
             // means the config line was found.
-            r.shell.is_wrapper_based()
-                || matches!(r.action, ConfigAction::AlreadyExists)
+            r.shell.is_wrapper_based() || matches!(r.action, ConfigAction::AlreadyExists)
         })
         .map(|r| r.path.as_path())
         .collect();


max-sixty:
@worktrunk-bot /tend-ci-runner:review 1736

max-sixty:
thanks @saschabratton . it does sound like our existing code had a problem

one concern is that the only tests that we have seem to revert the useful warnings.

@worktrunk-bot do you think you could
- check whether those tests were asserting good behavior
- if so, find out how restore the good behavior

and
- add comments explaining this behavior
- add tests for it

max-sixty:
@worktrunk-bot please can you give some simple examples of where we hit this false positive

if possible include those in tests but at least explain here

max-sixty:
excellent, thank you!

would there be a simpler way to design this? I worry a bit about the shell `match` sparse from the original check...

saschabratton:
thanks @max-sixty!

PR #1743, opened 2026-03-26
Add list ignore support for worktree paths and branch names

willpote:
Adds [list].ignore config. These are glob patterns that hide worktrees and branches from wt list output. Motivation is when using tools like Cursor, which generate worktrees for each agent running in parallel, `wt list` becomes cluttered with these temporary worktrees and un-usable.

```
  [list]
  ignore = ["tmp-*", "*/scratch/*"]
```
Per-project overrides work via the `[projects."..."]` table (inherits through OverridableConfig).

Each pattern is matched two ways:
  - As a path glob against the worktree's canonical filesystem path (e.g. */scratch/* matches worktrees inside a scratch/ directory)
  - As a string match against the branch name (e.g. tmp-* matches any branch starting with tmp-)

A worktree or branch is hidden if either matches. Filtering applies uniformly to worktrees, local branches (--branches), and remote branches (--remotes).

`wt list --ignored` bypasses all ignore filtering and shows everything.

When items are filtered, the summary line includes the count: Showing 3 worktrees, 1 ignored.

  Changes
  - ignore: Option<Vec<String>> field on ListConfig
  - --ignored CLI flag on wt list
  - Filtering in collect() after worktree/branch collection, before sorting — matches worktrees by path and branch name, local/remote branches by name
  - SummaryMetrics.ignored count shown in the summary line when > 0
  - glob crate added as a dependency
  - 6 integration tests: single glob, array globs, --ignored flag, parent path matching, branch name matching, --branches filtering


max-sixty:
I think I get the motivation, and I've had a similar experience at times

but I worry this might not quite pull its weight... 

hardly my style to suggest multiple checkouts, but one approach is to give cursor another base checkout

OTOH I can't think of an easy way to do this without the feature. and we could make the downside less confusing with "X hidden" in the footer

willpote:
> I think I get the motivation, and I've had a similar experience at times
> 
> but I worry this might not quite pull its weight...
> 
> hardly my style to suggest multiple checkouts, but one approach is to give cursor another base checkout
> 
> OTOH I can't think of an easy way to do this without the feature. and we could make the downside less confusing with "X hidden" in the footer

In the Cursor case, if you use the "Use multiple agents" feature when implementing something, it automatically spins up worktrees for you in `~/.cursor/worktrees`, its not something another base checkout helps with unfortunately

Re: X hidden, i did add that to the summary message- not sure if this is what you meant but it says `○ Showing 1 worktree, 4 ignored, 1 with changes`

max-sixty:
ok. let me think / let's see whether anyone else has thoughts

feel free to use @worktrunk-bot to fix tests etc

max-sixty:
> Re: X hidden, i did add that to the summary message- not sure if this is what you meant but it says ○ Showing 1 worktree, 4 ignored, 1 with changes

I was thinking "8 hidden" could be here

willpote:
> > Re: X hidden, i did add that to the summary message- not sure if this is what you meant but it says ○ Showing 1 worktree, 4 ignored, 1 with changes
> 
> I was thinking "8 hidden" could be here

Right i added 'X ignored'. I can rename to 'X hidden' if you like

max-sixty:
> Right i added 'X ignored'.

sorry, not sure how I missed that!

max-sixty:
@worktrunk-bot can we fix the tests, so we have a working PR to review?

willpote:
@worktrunk-bot update dev/config.example.toml as suggested

willpote:
@max-sixty any thoughts on this change?

max-sixty:
@worktrunk-bot do we label as experimental? can we do that if not?

thanks @willpote

max-sixty:
@worktrunk-bot can we review whether this should be `ignore` vs `hidden`? if we can switch to them etc, it should be "hidden"; it should only be "ignored" if we totally ignore it

max-sixty:
@worktrunk-bot let's make the naming change and then if @willpote disagrees we can revert

max-sixty:
@worktrunk-bot merge and resolve conflicts

max-sixty:
let's remove `--hidden`; I'm not sure exactly what that should be called; I'm not sure it should even be an option; and we can add later. Add a TODO re adding this option later

max-sixty:
@willpote is this version OK with you?

max-sixty:
closing as stale but open to adding this if folks would like it. feel free to ask @worktrunk-bot to re-implement if so

PR #1771, opened 2026-03-27
Fall back to effective URL for forge detection with url.insteadOf

amodelaweb:
When users configure `url.insteadOf` for multi-key SSH setups (e.g., `git@github-work:org/repo.git` rewritten to `git@github.com:org/repo.git`), the raw remote URL has a custom hostname that doesn't match any known forge. This breaks CI status, PR detection, and push-remote features.

The fix: add `effective_remote_url` which uses `git remote get-url` (applies `insteadOf` rewrites) and cache results per-remote in `RepoCache`. This is a local operation (no network I/O). When no `insteadOf` rules are configured, the effective URL equals the raw URL — no behavioral change. Both `remote_url` (raw) and `effective_remote_url` (rewritten) are needed because `insteadOf` can be configured in either orientation relative to forge URLs — forward (custom hostname → real forge, e.g. multi-key SSH) or reverse (real forge → internal destination, e.g. corporate mirrors).

Key changes:
- `effective_remote_url` cached per-remote via `DashMap` — free after first call
- Forge detection callers (`find_forge_remote`, `platform_for_repo`, `github_push_url`, `detect_github`, `fetch_pr_info`) use `effective_remote_url`
- `find_remote_for_repo` checks both raw and effective URLs (needs both for forward and reverse `insteadOf` scenarios)
- Removed `forge_remote_url`, `is_known_forge`, `primary_forge_remote_url` — superseded by the cached `effective_remote_url`
- Fixed `detect_github` using raw `remote_url` for remote branch owner lookup

> _This was written by Claude Code on behalf of amodelaweb_

### DIFF
--- a/src/git/repository/remotes.rs
+++ b/src/git/repository/remotes.rs
@@ -24,7 +24,6 @@
 //! - `effective_remote_url` — `git remote get-url`, with `insteadOf` applied
 //!   (cached; use for all forge detection)
 //! - `find_forge_remote(pred)` — search all remotes using effective URLs
-//! - `primary_effective_remote_url` — convenience for the primary remote
 
 use anyhow::Context;
 
@@ -228,16 +227,6 @@ impl Repository {
             }
         }
         None
-    }
-
-    /// Get the effective URL for the primary remote.
-    ///
-    /// Convenience for `primary_remote()` + `effective_remote_url()`. Applies
-    /// `url.insteadOf` rewrites, so custom SSH hostnames resolve to real forge
-    /// hostnames.
-    pub fn primary_effective_remote_url(&self) -> Option<String> {
-        let remote = self.primary_remote().ok()?;
-        self.effective_remote_url(&remote)
     }
 
     /// Get a project identifier for approval tracking.
--- a/src/git/remote_ref/github.rs
+++ b/src/git/remote_ref/github.rs
@@ -84,7 +84,9 @@ fn fetch_pr_info(pr_number: u32, repo: &Repository) -> anyhow::Result<RemoteRefI
     // hostnames (e.g., github-work → github.com for multi-key SSH setups).
     // Falls back to gh's default (github.com) if the remote URL can't be parsed.
     let hostname = repo
-        .primary_effective_remote_url()
+        .primary_remote()
+        .ok()
+        .and_then(|remote| repo.effective_remote_url(&remote))
         .and_then(|url| git::GitRemoteUrl::parse(&url))
         .map(|parsed| parsed.host().to_string())
         .unwrap_or_else(|| "github.com".to_string());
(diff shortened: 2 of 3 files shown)


max-sixty:
is setting the type of forge in the configs an alternative?

does this require an additional network request, and if so, when?

@amodelaweb if we did go ahead, we would need to get the test coverage much higher, potentially by consolidating some of the code

amodelaweb:
Thanks for the review @max-sixty !

**Config alternative:** Yeah, I think both approaches have merit. The config approach (`forge = "github"`) is simpler and explicit — no magic. The insteadOf approach is zero-config but adds complexity. One middle ground: we could support both, with config taking precedence (like `[ci] platform` already does). But if you'd prefer config-only, I'm happy to simplify this PR to just add a `forge` config option and skip the insteadOf inference entirely. What's your preference?

**Network requests:** Zero — `git remote get-url` is purely local. It reads `.git/config` and applies `url.insteadOf` rewrites in-process. The only cost is one extra `git` subprocess spawn on the fallback path, and only when the raw hostname isn't recognized (so no impact for the common case).

**Test coverage:** Agreed. I've already consolidated the two-pass iteration into `Repository::find_forge_remote()` in the latest push, which makes the core logic directly testable. The remaining coverage gaps are in the thin consumer wrappers that delegate to it. I'll keep pushing on this.

Let me know which direction you'd like to go on the config vs insteadOf question — happy to adjust.

max-sixty:
@worktrunk-bot please make those changes to improve the coverage

@amodelaweb thanks for the PR; please don't paste AI output from your handle. we have a bot for that!

max-sixty:
@worktrunk-bot please do a big consolidation, I see a bunch of repetition. run /simplify

max-sixty:
@worktrunk-bot can we write a spec at the top of a module of exactly how we resolve what forge to use?

src/git/repository/remotes.rs
```diff
@@ -1,4 +1,42 @@
...
+//!    hostname is a known forge (contains "github" or "gitlab"), use it directly.
+//!
+//! 2. **insteadOf fallback** — if the hostname is unrecognized, call
+//!    `git remote get-url <name>` which applies `url.insteadOf` rewrites to
+//!    produce the effective URL. If *that* hostname is a known forge, use the
+//!    effective URL instead. This is a local operation (no network I/O), costing
+//!    only one extra git subprocess.
+//!
+//! 3. **Best effort** — if neither URL resolves to a known forge, return the
+//!    raw URL so callers can still attempt to parse owner/repo from it.
```
max-sixty: Why have logic in the caller? why not have all the logic in this one function, and otherwise fail?

max-sixty:
@worktrunk-bot note my question above

PR #1864, opened 2026-04-01
Fix/copy ignored too many open files

### LINKED ISSUE #1865: copy-ignored fails silently with very large trees
I noticed today that I was not getting `node_modules` copied over correctly when creating new worktrees.

My `~/.config/worktrunk/config.toml` only contains the following at the end of the default file:

```toml
[post-start]
copy = "wt step copy-ignored"
```

When I ran `wt switch --create`, I got what appeared to be a correct worktree and no error messages, but some of my `npm` packages were not there. These are not git-controlled files.

Next, I ran `wt hook post-start -v` to debug and got:

```bash
$ wt hook post-start -v
○ Expanding user:copy
  wt step copy-ignored
  →
  wt step copy-ignored
◎ Running post-start user:copy
  wt step copy-ignored
```

Nothing obviously wrong there. Then I ran `wt step copy-ignored -v`:

```bash
$ wt step copy-ignored -v
○ Copying 2628 entries:
  .env (file)
<snip>
tmp/storage/zw (dir)
  tmp/storage/zx (dir)
  tmp/storage/zy (dir)
  tmp/storage/zz (dir)
✗ copying directory storage/04
  Too many open files (os error 24)
```

This is on macOS 26.4. It's a Rails project with a large seeded storage directory and a sub-project with a very large `node_modules` directory.

I can probably work around this by scripting a fresh `rails db:seed` and `npm ci` as a hook but the bigger issue, I think, is that an error in a low-level step isn't being propagated to the user.

While I think that rewriting the background processes to properly report errors is a bigger issue, https://github.com/max-sixty/worktrunk/pull/1864 is my effort to address the underlying failure.

fspeirs:
This PR fixes a problem when copying large `.gitignored` trees between worktrees.

The underlying problem is that each iteration of `entries_to_copy` created its own Rayon thread pool each with `MAX_COPY_THREADS` which eventually exhausted the number of possible open files on my macOS 26.4 installation.

I uncovered this when trying to create worktrees with very large gitignored subtrees.

The fix changes `step_copy_ignored` to create a 4-thread pool and have `copy_dir_recursive` check if it's being called from a Rayon thread context and skip creating its own thread pool if that is the case.

Closes #1865

### DIFF
--- a/tests/integration_tests/step_copy_ignored.rs
+++ b/tests/integration_tests/step_copy_ignored.rs
@@ -1390,7 +1390,11 @@ fn test_copy_ignored_many_directories_no_emfile(mut repo: TestRepo) {
         let dir = repo.root_path().join(format!("ignored-dir-{i}"));
         fs::create_dir_all(&dir).unwrap();
         for j in 0..10 {
-            fs::write(dir.join(format!("file-{j}.txt")), format!("content {i}-{j}")).unwrap();
+            fs::write(
+                dir.join(format!("file-{j}.txt")),
+                format!("content {i}-{j}"),
+            )
+            .unwrap();
         }
         gitignore_entries.push_str(&dir_name);
         gitignore_entries.push('\n');


src/copy.rs
```diff
@@ -136,3 +147,71 @@ pub fn create_symlink(target: &Path, src_path: &Path, dest_path: &Path) -> anyho
...
+        pool.install(|| {
+            (0..DIR_COUNT)
+                .into_par_iter()
+                .try_for_each(|i| -> anyhow::Result<()> {
+                    let src = src_root.path().join(format!("dir-{i}"));
+                    let dst = dst_root.path().join(format!("dir-{i}"));
+                    copy_dir_recursive(&src, &dst, false)
+                })
+        })
+        .unwrap();
```
max-sixty: we need to test our actual code; not a duplicate of our code

@worktrunk-bot can you suggest some options?

max-sixty:
@worktrunk-bot note the request above

max-sixty:
@worktrunk-bot do request #1 from above

(unless @fspeirs objects / has other views)

fspeirs:
@max-sixty That's exactly what I thought about this morning! Happy with that.

PR #1907, opened 2026-04-04
docs: add cmux workspace integration recipe

alvistar:
## Summary

- Adds cmux to the **Agent handoffs** section alongside tmux and Zellij
- Adds a dedicated **cmux workspace per worktree** recipe with create/select/close lifecycle hooks
- Documents the key gotcha: cmux socket restricts access to processes with cmux terminal ancestry, so `pre-*` hooks must be used instead of `post-*`

## Context

[cmux](https://cmux.com) is a macOS terminal built on libghostty with workspace management via a Unix socket API. This recipe wires up worktrunk hooks so each worktree automatically gets its own cmux workspace.

The `pre-*` vs `post-*` distinction was discovered through debugging: `post-*` hooks are detached background processes that lose the cmux process ancestry, causing the socket to reject connections with "Access denied — only processes started inside cmux can connect."

## Test plan

- [x] Tested full lifecycle: `wt switch -c`, `wt switch`, `wt remove` — all correctly create, select, and close cmux workspaces
- [ ] Verify docs render correctly on worktrunk.dev (Zola build)


docs/content/tips-patterns.md
```diff
@@ -1,6 +1,6 @@
 +++
 title = "Tips & Patterns"
-description = "Practical recipes for Worktrunk workflows: aliases, shell integration, Zellij layouts, and parallel agent patterns."
+description = "Practical recipes for Worktrunk workflows: aliases, shell integration, Zellij layouts, cmux workspaces, and parallel agent patterns."
```
max-sixty: revert please

docs/content/tips-patterns.md
```diff
@@ -262,6 +265,32 @@ To create a worktree and immediately attach:
...
 {% end %}
 
+## cmux workspace per worktree
+
+Each worktree gets its own [cmux](https://cmux.com) workspace. Switching worktrees switches workspaces; removing a worktree closes its workspace.
+
+```toml
+# ~/.config/worktrunk/config.toml
+[pre-start]
+cmux = "cmux new-workspace --name {{ repo | sanitize }}/{{ branch | sanitize }} --cwd {{ worktree_path }}"
```
max-sixty, line 330: @alvistar are you very confident these are correct and work well?

max-sixty (approved):
thanks @alvistar !

docs/content/tips-patterns.md
```diff
@@ -262,6 +265,32 @@ To create a worktree and immediately attach:
...
+
+[pre-switch]
+cmux = """
+WS=$(cmux list-workspaces 2>/dev/null | grep -F {{ repo | sanitize }}/{{ branch | sanitize }} | head -1 | grep -o 'workspace:[0-9]*')
+[ -n "$WS" ] && cmux select-workspace --workspace "$WS" || true
+"""
+
+[pre-remove]
+cmux = """
```
kay-schecker, line 341: The `grep -F` here is a substring match — it resolves to the wrong workspace when two workspace names share a prefix.

`cmux list-workspaces` prints one per line:

```
  workspace:6 web/demo2
* workspace:5 web/demo  [selected]
```

With `branch=demo`, `grep -F web/demo` matches both lines and `head -1` returns `workspace:6`, so `pre-switch`/`pre-remove` select or close `demo2` instead of `demo` — silently operating on the wrong worktree. The `pre-remove` case is especially bad: closing the wrong workspace.

An exact field match avoids that:

```bash
WS=$(cmux list-workspaces 2>/dev/null | awk -v name="{{ repo | sanitize }}/{{ branch | sanitize }}" \
  '{for (i=1;i<=NF;i++) if ($i ~ /^workspace:[0-9]+$/ && $(i+1)==name) {print $i; exit}}')
```

`awk` finds the `workspace:N` token and compares the next field exactly to `repo/branch`. That also makes the trailing explanation about the `*` prefix unnecessary, since `awk` handles the selected-line layout naturally.

skills/worktrunk/reference/tips-patterns.md
```diff
@@ -279,6 +285,32 @@ To create a worktree and immediately attach:
...
+cmux = "cmux new-workspace --name {{ repo | sanitize }}/{{ branch | sanitize }} --cwd {{ worktree_path }}"
+
+[pre-switch]
+cmux = """
+WS=$(cmux list-workspaces 2>/dev/null | grep -F {{ repo | sanitize }}/{{ branch | sanitize }} | head -1 | grep -o 'workspace:[0-9]*')
+[ -n "$WS" ] && cmux select-workspace --workspace "$WS" || true
+"""
+
+[pre-remove]
+cmux = """
```
kay-schecker, line 361: Same issue as in `docs/content/tips-patterns.md`: `grep -F` is a substring match, so with workspaces like `repo/demo` and `repo/demo2`, `branch=demo` matches both lines and `head -1` returns the wrong `workspace:N`. `pre-remove` then closes the wrong workspace.

Suggested exact-field match via `awk`:

```bash
WS=$(cmux list-workspaces 2>/dev/null | awk -v name="{{ repo | sanitize }}/{{ branch | sanitize }}" \
  '{for (i=1;i<=NF;i++) if ($i ~ /^workspace:[0-9]+$/ && $(i+1)==name) {print $i; exit}}')
```

This also obviates the trailing note about the `*` prefix, since `awk` handles the selected-line layout naturally.

kay-schecker (commented):
Heads-up: the `grep -F` match in the `pre-switch`/`pre-remove` snippets is a substring match, so sibling workspaces that share a prefix (`demo` / `demo2`) resolve to the wrong `workspace:N`. Details + an `awk`-based fix inline.

max-sixty:
Does `cmux` have a structured output, so we can do less string matching?

@worktrunk-bot resolve conflicts; integrate the suggested changes after verifying them, thanks

kay-schecker:
Bildschirmfoto 2026-04-23 um 14 58 30

@max-sixty there is no structured output in the latest version (0.63.2)

max-sixty:
@worktrunk-bot please revert the recent changes if that's the case... also please put an issue into the tend repo; possibly we should suggest testing changes manually which don't have automated tests

sorry @kay-schecker and thanks for pointing that out

tend-agent:
Following up from [tend#327](https://github.com/max-sixty/tend/pull/327#issuecomment-4311382996): `--json` *is* wired into `list-workspaces` in v0.63.2 — cmux just parses it as a global flag, before the subcommand.

The pre-command parser at [`CLI/cmux.swift:1635-1690`](https://github.com/manaflow-ai/cmux/blob/v0.63.2/CLI/cmux.swift#L1635-L1690) reads `--json` and then `break`s at the first non-global token. The handler at [L2003-2006](https://github.com/manaflow-ai/cmux/blob/v0.63.2/CLI/cmux.swift#L2003-L2006) honors `jsonOutput`. So:

- `cmux list-workspaces --json` → `--json` falls through to `commandArgs` and is silently ignored; text output. (The form the mintlify reference shows, and what I copied earlier.)
- `cmux --json list-workspaces` → `{"workspaces": [{"ref", "id", "index", "title", "selected", ...}]}`.

So the structured-output approach is salvageable with the flag moved before the subcommand.

max-sixty:
> @max-sixty want me to take another pass with `cmux --json list-workspaces` (and someone with cmux installed sanity-checking the output before merge),

yes, and you should be downloading cmux and checking!

docs/content/tips-patterns.md
```diff
@@ -315,6 +318,38 @@ To create a worktree and immediately attach:
...
+WS=$(cmux --json list-workspaces 2>/dev/null \\
+  | jq -r --arg t '{{ repo | sanitize }}/{{ branch | sanitize }}' \\
+      '.workspaces[] | select(.title == $t) | .ref' | head -1)
+[ -n "$WS" ] && cmux close-workspace --workspace "$WS" || true
+"""
+```
+
+**Why `pre-*` instead of `post-*`?** cmux restricts socket access to processes spawned inside a cmux terminal. `post-*` hooks run as detached background processes, breaking the process ancestry chain. `pre-*` hooks run in the foreground and inherit the terminal's process lineage.
+
+**Why `cmux --json` and not `cmux list-workspaces --json`?** cmux parses `--json` as a pre-command flag, before the subcommand — placed after the subcommand it falls through to the command's argument list and is silently ignored, leaving you with text output. Exact-matching on `.title` then avoids the substring bleed that `grep -F` would cause for sibling workspaces sharing a prefix (e.g., `demo` vs `demo2`).
```
max-sixty: @worktrunk-bot remove this line

docs/content/tips-patterns.md
max-sixty: @worktrunk-bot please make this change

max-sixty: we want to remove this change @worktrunk-bot

PR #1935, opened 2026-04-06
fix(opencode): fix broken unicode escaping depending on Bun version

noirbizarre:
A small fix on the OpenCode plugin: depending on the Bun version, unicode can be broken ([escaped while it shouldn't](https://github.com/bikeshaving/crank/issues/342)).

This fix ensures it works whatever the Bun version used in OpenCode (I already use this fix locally).


max-sixty:
thanks @noirbizarre !

PR #2064, opened 2026-04-10
Unify conflict-detection skip sets between wt list and wt switch

max-sixty:
Both `wt list` (non-full) and the `wt switch` picker now run both `WorkingTreeConflicts` and `MergeTreeConflicts`. The only remaining asymmetry is `SummaryGenerate`: `wt list` still skips it (doesn't render summaries inline), the picker still runs it (tab 5 preview).

## Why

The picker's old skip set was inconsistent in two ways:

1. **Cost inverted.** It skipped the cheaper probe (`MergeTreeConflicts` = one `merge-tree HEAD base` call) and ran the more expensive one (`WorkingTreeConflicts` = `git status` + `git stash create` + `git merge-tree` for dirty worktrees).

2. **Fallback unreachable.** `WorkingTreeConflicts` short-circuits to `None` on clean worktrees with an explicit contract that the caller falls back to `MergeTreeConflicts` (see `src/commands/list/collect/tasks.rs:437`). The picker blocked that fallback, so clean worktrees in the picker produced no real conflict answer — they fell through to the struct default.

`wt list` non-full's old choice was internally consistent (skip the working-tree probe, always use commit-level), but asymmetric with the picker. Adding `WorkingTreeConflicts` to `wt list` is strictly a correctness improvement: rows with a conflicted rebase-in-progress now render `✗` instead of a counts-based fallback.

## Safety

Adding `MergeTreeConflicts` to the picker is safe because the per-symbol atomic rendering refactor on this branch makes gate 3 position 4 render as `⋯` when its inputs haven't resolved within the collect deadline. Tail cost is capped, not unbounded.

## Reviewer tour

**Commit 1** (skip-set unification):
- `src/commands/list/collect/mod.rs` — one line removed from non-full skip set.
- `src/commands/picker/mod.rs` — one line removed from picker skip set; cargo fmt collapsed the now-2-element array.
- `src/commands/list/layout.rs` — test helper `non_full_skip_tasks()` updated to match the new production skip set (its doc comment claimed to mirror non-full mode).
- `tests/integration_tests/list.rs` — renamed `test_list_full_working_tree_conflicts` → `test_list_working_tree_conflicts` and updated its module doc + inline comments; the test now documents that both modes detect working-tree conflicts.
- Two snapshot updates are the point of the change — one gains `✗` for a real conflict, the other gains a second "task failed" warning line (both conflict tasks now fail symmetrically on a broken HEAD).

**Commit 2** (doc-comment cleanup, from auto-review feedback):
- Dropped stale "--full only" qualifiers from 5 doc-comment locations in `tasks.rs`, `types.rs`, and `mod.rs`, plus one in the test file. No behavior change.

## Follow-up worth considering separately

For dirty worktrees the picker now runs both `WorkingTreeConflicts` and `MergeTreeConflicts`, duplicating a `merge-tree` call. If that shows up in profiles, the right fix is conditional dispatch — skip `MergeTreeConflicts` when `WorkingTreeConflicts` produced a non-`None` answer — rather than reinstating the asymmetric skip sets.

### DIFF
--- a/src/commands/list/collect/mod.rs
+++ b/src/commands/list/collect/mod.rs
@@ -217,7 +217,6 @@ pub struct CollectOptions {
     ///   since they depend on skipped tasks
     ///
     /// Note: `wt switch` interactive picker doesn't show the BranchDiff column, so `…` isn't visible there.
-    /// This is similar to how `✗` conflict only shows with `--full` even in `wt list`.
     ///
     /// TODO: Consider adding a visible indicator in Status column when integration
     /// checks are skipped, so users know the `⊂` symbol may be incomplete.
--- a/src/commands/list/collect/mod.rs
+++ b/src/commands/list/collect/mod.rs
@@ -391,7 +391,6 @@ pub fn collect(
                     [
                         TaskKind::BranchDiff,
                         TaskKind::CiStatus,
-                        TaskKind::WorkingTreeConflicts,
                         TaskKind::SummaryGenerate,
                     ]
                     .into_iter()
--- a/src/commands/list/collect/tasks.rs
+++ b/src/commands/list/collect/tasks.rs
@@ -430,7 +430,7 @@ impl Task for MergeTreeConflictsTask {
     }
 }
 
-/// Task 6b (worktree only, --full only): Working tree conflict check
+/// Task 6b (worktree only): Working tree conflict check
 ///
 /// For dirty worktrees, uses `git stash create` to get a tree object that
 /// includes uncommitted changes, then runs merge-tree against that.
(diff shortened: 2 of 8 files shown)


max-sixty:
Thanks for the catch. Fixed in c01898b7b — dropped the stale "--full only" qualifiers from the 5 doc locations you flagged (`tasks.rs`, `types.rs` ×3, `mod.rs` stale-branches analogy), plus one more I found while grepping: `tests/integration_tests/list.rs:3066` listed `WorkingTreeConflicts` as a "--full enables" example. No behavior change.

PR #2100, opened 2026-04-11
docs(config): drop stale approved-commands example

max-sixty:
The deprecation layer migrates `approved-commands` out of the user config file into `approvals.toml` on load, so showing it inside a `[projects.*]` block in the `UserProjectOverrides` doc comment is misleading. One-line removal.


max-sixty:
Good catch — fixed in 44e98b40b. Replaced the per-project block with a `worktree-path` override so it still illustrates something real.

PR #2130, opened 2026-04-12
Run internal detached ops at nice 19

max-sixty:
`wt remove`'s background `rm -rf` and `sweep_stale_trash` were competing with foreground work for CPU/IO. This extends the renicing already done for `step_copy_ignored` (see `src/copy.rs::lower_process_priority`) to the detached-spawn paths: `spawn_detached` and `spawn_detached_exec` now wrap the child in `nice -n 19 --` when `hook_log` is `HookLog::Internal`. User hooks keep running at normal priority.

The nice value is inherited across fork/exec, so wrapping the outer `sh` (for the shell-based path) or the target binary directly (for the exec path) is enough — the backgrounded work inherits it.

`nice` is POSIX and shipped in base macOS/Linux; a missing binary would fail the spawn, matching the existing `renice` assumption in `copy.rs`. Windows is left on its own priority mechanism (not wired up here).

A follow-up is under investigation: macOS has `taskpolicy -b` which also throttles I/O (IOPOL_THROTTLE) rather than just CPU priority. If the research lands in favor of it, the single `low_priority_command` helper is the one place to swap the primitive.


max-sixty (commented):
Added the helper test as suggested. Thanks for catching the coverage gap.

PR #2185, opened 2026-04-13
fix(nix): update flake lock so Nix build meets Rust 1.93 MSRV

Lysanleo:
## Problem

After the recent minimal supported Rust version bump to Rust 1.93, the Nix dev/build path is still pulling in Rust 1.92 from the locked `rust-overlay`, so `nix develop -c cargo build` fails with:  `worktrunk requires rustc 1.93`

## Cause

  `Cargo.toml` now requires Rust 1.93, but `flake.lock` still points to an older `rust-overlay` revision, and `flake.nix` uses `pkgs.rust-bin.stable.latest.default`.

## Fix

Refresh the locked `rust-overlay` input so the flake provides a Rust toolchain that satisfies the current MSRV.

## Validation

Run:

- `nix build .#worktrunk`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -64,11 +64,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1767926800,
-        "narHash": "sha256-x0n73J6ufD/EhDlVdcoAmF0OQHZ+b0a2cKDc8RZyt+o=",
+        "lastModified": 1776050130,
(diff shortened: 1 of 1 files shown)


max-sixty:
@worktrunk-bot how can we keep these more in sync in the future?

max-sixty:
great, open a PR for those two

PR #2187, opened 2026-04-13
Fix formatting in installation commands in README

MahmoudMabrok:
removing `$ ` as it not part of command, current state make copy button in readme to copy it then throw error like `zsh: command not found: $`


max-sixty:
@worktrunk-bot could you look into this more

let's indeed keep for the console blocks

for the bash blocks, how feasible is removing those? it would be marginally better

PR #2240, opened 2026-04-15
test(picker): remove dead disable_picker_timeout helper

max-sixty:
The helper in `tests/integration_tests/switch_picker.rs` wrote to top-level `[switch-picker]` (with a hyphen), which the config loader has always silently ignored — the real section is `[switch.picker]`. On top of that, progressive rendering (#2231) removed the picker's reliance on any timeout, and `switch.picker.timeout-ms` was formally deprecated in #2236.

So the helper has been a no-op for multiple reasons. Removed the function and all 7 call sites. Picker tests still pass without it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Changelog
 
+## 0.37.1
+
+### Improved
+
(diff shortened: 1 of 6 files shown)


max-sixty:
yes please

PR #2258, opened 2026-04-16
docs(perf): require --progressive in wt-perf trace pipeline and trim duplication

max-sixty:
Two TTY-gated trace events (`Skeleton rendered`, `First result received`) don't fire when `wt list` detects a non-TTY stdout, which every piped `wt-perf` invocation does. Every documented capture pipeline now passes `--progressive` so all seven instant events flow through `wt-perf trace` / `cache-check`. Fixes 10 sites across module docs, clap `after_long_help`, setup eprintln, and `benches/CLAUDE.md`.

While in the area, I also consolidated some of the scattered examples:

- Dropped the `# CLI Usage` block from `tests/helpers/wt-perf/src/lib.rs` — library docs shouldn't carry CLI examples, and they fully duplicate `wt-perf --help`.
- Trimmed the `main.rs` module docstring from four example invocations to a one-liner pointing at `wt-perf --help`.
- Dropped the "with benchmark repo" variant from `cache-check`'s `after_long_help`, since it's structurally identical to the first example with `-C <path>` prepended.
- Replaced the stdin-hint and no-entries error in `wt-perf` (which repeated the full pipeline) with a pointer to `<subcommand> --help`.
- Pruned `src/trace/chrome.rs`'s standalone usage example to defer to `crate::trace`.

Verified end-to-end: `RUST_LOG=debug wt list --progressive 2>&1 | wt-perf trace` emits all seven expected instant events in the Chrome Trace JSON.

### DIFF
--- a/benches/CLAUDE.md
+++ b/benches/CLAUDE.md
@@ -198,7 +198,7 @@ trace_processor trace.json -q /tmp/q.sql
 
 ```bash
 # Trace on rust-lang/rust (must run benchmark first to clone)
-RUST_LOG=debug cargo run --release -q -- -C target/bench-repos/rust list --branches 2>&1 \
+RUST_LOG=debug cargo run --release -q -- -C target/bench-repos/rust list --progressive --branches 2>&1 \
   | cargo run -p wt-perf -- trace > rust-trace.json
 ```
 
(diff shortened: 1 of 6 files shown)


max-sixty:
Thanks — good catch. Applied in 8b988c0. Also caught a test regression from the earlier error-message consolidation (`test_wt_perf_trace_empty_input` was asserting the old "No trace entries found" wording) and fixed that in the same commit.

PR #2276, opened 2026-04-17
fix: encode `/` in branch names for git config keys

Mziserman:
## Problem

`wt config state vars set key=value` fails for branches containing `/` (e.g. `feature/my-task`):

```
✗ Invalid key "worktrunk.state.feature/my-task.vars.key": ...
```

Git config rejects `/` in key path segments, so any slash-branch silently broke the entire vars feature.

## Solution

Encode `/` as `..` when writing branch names into git config keys, decode on read.

**Why `..`?** Git forbids `..` in branch names via `git-check-ref-format`, making it a collision-free, reversible encoding. Other safe-in-config characters (`-`, alphanumeric) are also valid in branch names and would cause collisions.

Also tightens `validate_vars_key` to reject `_`, which git config likewise forbids in variable names (last segment must match `[a-zA-Z][a-zA-Z0-9-]*`).

## Changes

- `src/git/repository/config.rs`: encode branch in `branch_marker`, `vars_entries`, `all_vars_entries`
- `src/commands/config/state.rs`: encode branch in all `worktrunk.state.*` key constructions; decode when reading back in `all_markers`
- `tests/integration_tests/config_state.rs`: new `test_vars_slash_branch` test; updated `test_vars_invalid_key` snapshot

### DIFF
--- a/src/commands/config/state.rs
+++ b/src/commands/config/state.rs
@@ -560,7 +560,10 @@ pub fn handle_state_get(
             };
             if format == SwitchFormat::Json {
                 // Read raw config to get both marker and set_at
-                let config_key = format!("worktrunk.state.{branch_name}.marker");
+                let config_key = format!(
+                    "worktrunk.state.{}.marker",
+                    encode_branch_for_config(&branch_name)
+                );
                 let raw = repo
                     .run_command(&["config", "--get", &config_key])
                     .ok()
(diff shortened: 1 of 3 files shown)


max-sixty:
@worktrunk-bot is this a bug? I see that @Mziserman closed the PR, but if this is a real bug then we should definitely fix

(thanks @Mziserman !)

Mziserman:
@max-sixty i was incorrect in the source of the error but i think there is an issue : 
```
  wt config state vars set db_suffix={{ branch | sanitize_db }}
  store │ ✗ error: invalid key: worktrunk.state.test-log2.vars.db_suffix
```
the worktrunk validator doesn't match git's validator but it's not about / in branch names it's about _ in variables. I'll open another pr.

PR #2285, opened 2026-04-18
fix: reject underscore in vars keys (invalid git config variable name)

Mziserman:
## Problem

`wt config state vars set db_suffix=foo` silently passes validation but then fails with a cryptic git error:

```
error: invalid key: worktrunk.state.main.vars.db_suffix
```

Git config variable names (the last `.`-separated segment) must match `[a-zA-Z][a-zA-Z0-9-]*` — underscores are not allowed at that position.

## Fix

Reject `_` in `validate_vars_key` early, with a clear message pointing users to use hyphens instead:

```
Invalid key "db_suffix": keys must contain only letters, digits, and hyphens
```

No other changes — branch names with `/` already work correctly via git's quoted subsection form.

### DIFF
--- a/docs/content/config.md
+++ b/docs/content/config.md
@@ -1001,7 +1001,7 @@ dev = "npm start -- --port {{ vars.config.port }}"
 
 ### Storage format
 
-Stored in git config as `worktrunk.state.<branch>.vars.<key>`. Keys must contain only letters, digits, hyphens, and underscores — dots conflict with git config's section separator.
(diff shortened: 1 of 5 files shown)


max-sixty:
thanks @Mziserman !

max-sixty:
@worktrunk-bot thanks for the review

in this case, I think you could have been more helpful by adding suggestions or at least offering to fix. could you put an issue into the tend repo for this case, and we can look at adjusting the guidance slightly?

PR #2296, opened 2026-04-18
feat(cli): reject alias args unused by template

max-sixty:
Closes a silent-drop footgun: `wt deploy production` with `deploy = "echo hello"` discarded `production` with no feedback; same for `wt deploy --env=prod` when the template had no `{{ env }}`. Both now error at the edge.

The check computes the union of referenced template variables across every command in the alias pipeline via `minijinja::Environment::template_from_str().undeclared_variables(false)` — the same AST-based primitive already used in `template_references_var`, `deprecation.rs`, and `hook_commands.rs`. One step referencing `{{ args }}` is enough to accept positionals for the whole pipeline. Templates that fail to parse skip the check entirely so the real syntax error still surfaces during expansion.

Runs before both the approval prompt and the `--dry-run` branch, so dry-run previews also fail on unused input — silently-dropped values wouldn't be visible in the preview anyway.

Secondary: `wt deploy -y` with no `{{ args }}` used to silently forward `-y` as a positional token; it now errors, pointing toward the `wt -y deploy` form the parser docstring already documents.

### Tests

Seven new integration tests in `tests/integration_tests/step_alias.rs` covering the three rejection cases (positional, single `--key=value`, multiple `--key=value`), post-alias `-y`, multi-step pipeline rejection, single-step pipeline union acceptance, and conditional `{% if args %}` references.

### DIFF
--- a/docs/content/extending.md
+++ b/docs/content/extending.md
@@ -122,6 +122,8 @@ open = "open http://localhost:{{ branch | hash_port }}"
 
 Hyphens in variable names are canonicalized to underscores at parse time, so `--my-var=value` is referenced as `{{ my_var }}` in templates. This lets flags use natural kebab-case while avoiding the minijinja parser's interpretation of `{{ my-var }}` as subtraction.
 
+Assignments for variables that no template step references are rejected — `wt deploy --env=prod` errors when no template in `deploy` uses `{{ env }}`, rather than silently discarding the value.
+
 ### Forwarding positional arguments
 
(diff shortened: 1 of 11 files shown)


max-sixty:
Superseded by #2287, which landed on main during development. #2287 took the 'route, don't reject' approach to the same problem space — unreferenced `--key=value` forwards to `{{ args }}` rather than erroring. Closing this PR as obsolete.

Issue #2322, opened 2026-04-20
aliases for internal subcommands seem to double runtime


markjaquith:
I'm using the new aliases feature for `rm`, `ls`, and `sw` to map to internal worktrunk subcommands.

These aliases consistently seem to nearly double the runtime:

## Normal `wt list`
```
❯ time (wt list)
  Branch  Status        HEAD±    main↕  Remote⇅  Path  Commit    Age   Message
@ main    ·  ·^··           ·                 ·  .     d2c72f2a  ·     ·

○ Showing 1 worktree
( wt list; )  1.30s user 4.17s system 220% cpu 2.474 total
```

## Slower `wt ls` alias
```
❯ time (wt ls)
◎ Running alias ls
  Branch  Status        HEAD±    main↕  Remote⇅  Path  Commit    Age   Message
@ main    ·  ·^··           ·                 ·  .     d2c72f2a  ·     ·

○ Showing 1 worktree
( wt ls; )  2.10s user 5.45s system 166% cpu 4.537 total
```

I believe what's happening is that there actually is double invocation, because the alias system is flexible enough to run ANYTHING.

I think there should be a fast code path for **internal** aliases which just are there to save keystrokes or map to people's mental models about ("`ParentCommand rm` is how I remove something"). It would also be nice to omit the "Running alias" message for these internal aliases.

max-sixty:
but @worktrunk-bot that should explain maybe tens of ms. Not 800ms! How do we get 800ms slower?

at least try to reproduce, please

max-sixty:
@markjaquith locally I happen to get 80ms: `wt list = 390ms, wt ls = 470ms`

That's a bit slow but no disaster — OTOH 800ms is much much worse.

Would you be able to share a diagnostic gist, by running `-vv` and running the printed command to create a secret gist, and then sharing that gist?

max-sixty:
there is one command that scales with the number of branches; I'm going to replace it with a different one which only looks at the current branch. let's see whether that helps

max-sixty:
@markjaquith would you mind testing out 0.41?

my low-confidence guess is that your repo has a lot of worktrees/branches. One of the queries into git is much faster running batch (during `wt list`) *but* we were using it for getting info on a single single worktree, which then is much slower. So I added a `_single` variant, which should be much faster. I'm getting 80ms performance now for 100 branches (but it may be a systems difference / it may be something else)

(and I think we can cut the 80ms too, for templates that don't use all the vars...)

[...]

PR #2339, opened 2026-04-20
Silence alias banner when there is no pipeline summary to show

max-sixty:
## Summary

Alias announcement banner (`◎ Running alias <name>`) now only prints when the alias has at least one named step to summarize. Single unnamed aliases (`ls = "wt list"`) and all-anonymous pipelines return no announcement — the banner was just echoing the user's typed name. Pipelines with named steps keep their informative summary (`Running alias deploy: install; build, lint`). `-v` still prints the bare form as a confirmation line.

Addresses half of #2322 (the other half — perf — landed in 4f9bd575a).

## Rule

- `[aliases] ls = "wt list"` → silent
- `[aliases] foo = ["a", "b"]` (anonymous pipeline) → silent
- `[aliases.check] build = "…", test = "…"` (concurrent named) → announce
- `[[aliases.deploy]] install = "…"` (named pipeline) → announce
- Any of the above under `-v` → announce (bare form if no summary)

Implementation: `format_alias_announcement` returns `Option<String>` — `None` when the summary would be empty. Caller falls back to the bare form only under `-v`.

## Alternatives considered

- **A. Auto-detect `wt <builtin>` passthrough.** Parse the template and silence only when it resolves to a trivial `wt <subcommand>` / `wt <subcommand> {{ args }}`. Narrower than the chosen approach — would still announce `deploy = "cd {{ worktree_path }} && make"`. Rejected: if the banner carries no summary, announcing it adds nothing regardless of what the template does; a token parser and `TOP_LEVEL_BUILTINS` coupling are extra machinery for no gain.
- **B. Explicit per-alias `quiet = true`.** Requires extending the schema: the current `[aliases.foo] k = "v"` form means "concurrent step named k", so reserving keys collides with real step names. Either a breaking reserved-key convention or a sibling `[alias-options.foo]` section. Rejected: extra user-visible knob for a default the data already implies.
- **C. Verbosity-gated (banner only at `-v`).** Silences pipelines too, even ones whose step summary is useful. Rejected: loses real signal for named pipelines.
- **D. Always silent (delete the banner).** Loses the informative summary case entirely. Rejected as baseline.
- **E. Global opt-out at top of user config.** Blunt — the reporter's case mixed trivial keystroke-savers with real pipelines, and a single global switch can't distinguish them. Rejected.

## Test plan

- [x] Unit tests: `test_format_alias_announcement_*` updated to assert `None` for single/anonymous cases, `Some(summary)` for named cases.
- [x] Integration snapshots: 16 single-step alias snapshots lose the banner line; `alias_pipeline_announcement`, `step_alias_multi_step_binds_across_pipeline` keep their summary; `alias_verbose_*` (both use `-v`) keep the bare banner via the verbosity branch.
- [x] `cargo run -- hook pre-merge --yes` — all 3296 tests pass, all lints pass.

### DIFF
--- a/src/commands/alias.rs
+++ b/src/commands/alias.rs
@@ -289,26 +289,28 @@ fn unknown_step_command_error(name: &str, alias_names: &[String]) -> anyhow::Err
     crate::enhance_clap_error(build_invalid_subcommand_error(step_cmd, name, suggestions))
 }
 
-/// Format the "Running alias …" announcement.
+/// Format the informative "Running alias …" announcement, or `None` when
+/// there is nothing beyond the alias name to announce.
 ///
-/// For pipelines or concurrent groups with named commands, includes a summary
-/// of the structure (e.g., `Running alias deploy: install; build, lint`).
-/// When all commands are unnamed, falls back to the bare form
-/// (`Running alias deploy`) — aliases have no natural fallback label for
-/// unnamed steps the way hooks use `user`/`project`.
+/// Returns `Some(…)` only when the alias has at least one named step, so the
+/// banner can carry a pipeline summary (e.g., `Running alias deploy: install;
+/// build, lint`). Single unnamed aliases (`ls = "wt list"`) and all-anonymous
+/// pipelines (`foo = ["a", "b"]`) return `None` — the banner would only echo
+/// the name the user just typed. Callers that still want a confirmation line
+/// (e.g. under `-v`) emit a bare `Running alias <name>` themselves.
 ///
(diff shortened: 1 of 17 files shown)


max-sixty:
@markjaquith WDYT of this for your suggestion " It would also be nice to omit the "Running alias" message for these internal aliases." — I agree small things should we nice and quiet

not set on this option if you or others have views. wanted to keep it simple / avoid adding configs etc

markjaquith:
I think that's a good rubric! Not worth specific configuration, and it aligns well with "simple alias = simpler output"

PR #2346, opened 2026-04-21
perf(config): route fsmonitor and pager reads through bulk cache

max-sixty:
Follow-up to #2344: the two `git config` subprocess call sites outside `RepoCache` that were intentionally left behind now consult the bulk `all_config` map.

`is_builtin_fsmonitor_enabled` reads `core.fsmonitor` via `config_last` and parses with `parse_git_bool`, matching how git itself routes the bool-or-string value — the builtin daemon is enabled for any git-bool truthy value, and anything else (including Watchman hook paths) counts as a hook/disabled. The practical widening over the previous `s.trim() == "true"` check is that `1`/`yes`/`on` now count as enabled; Watchman hook paths still correctly return false. A direct unit test seeds `all_config` on a fresh `RepoCache` and pins both the bool widening and the Watchman-stays-false invariant — the function's only production caller is macOS-gated, so the Linux coverage runner would otherwise see the body as dead code.

`git_config_pager` already built a `Repository::current()`, so the swap to `config_value` was one line.

Not migrated: `effective_remote_url` still calls `git remote get-url` because it needs `url.insteadOf` URL rewrites, which aren't a plain config lookup. It stays cached separately in `RepoCache.effective_remote_urls`.

### DIFF
--- a/src/git/repository/mod.rs
+++ b/src/git/repository/mod.rs
@@ -818,12 +818,16 @@ impl Repository {
 
     /// Check if git's builtin fsmonitor daemon is enabled.
     ///
-    /// Returns true only for `core.fsmonitor=true` (the builtin daemon).
-    /// Returns false for Watchman hooks, disabled, or unset.
+    /// Returns true for any git-bool truthy value (`true/1/yes/on`), which
+    /// matches how git itself routes the bool-or-string `core.fsmonitor`
+    /// config to the builtin daemon. Returns false for Watchman hook paths,
+    /// disabled, or unset.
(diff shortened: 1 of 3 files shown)


max-sixty:
Added the suggested unit test in 25ea95a47 — pre-populates `all_config` on a fresh `RepoCache` (per the pattern in `repo_path_error_when_is_bare_fails`) and exercises truthy git-bool values, the Watchman hook path, explicit falsy values, and unset. Pins the bool-widening intent and the Watchman-stays-false invariant.

Re the windows flake (`test_prune_json_orphan_branch`): rerun is green, agree it's unrelated.

PR #2350, opened 2026-04-21
perf(repo): skip submodule rev-parse probe via core.worktree in bulk config

max-sixty:
## Summary

Replaces the failing `git rev-parse --show-toplevel` probe in `Repository::repo_path()` with a direct read of `core.worktree` from the bulk config map. On non-submodule repos the probe fails unconditionally, costing roughly 5ms per call on healthy macOS — or ~200ms on systems with expensive `execve` like the one reported in #2322. `repo_path()` fires ~2× during parent-side alias dispatch, so the win compounds.

The new control flow matches git's own rules for locating the working tree:

| `git_common_dir` layout     | Signal                     | Resolution                       |
|-----------------------------|----------------------------|----------------------------------|
| Bare `.git`                 | `core.bare = true`         | `git_common_dir` is the repo     |
| Submodule `.git/modules/X`  | `core.worktree` set by git | join + normalize vs common dir   |
| Normal `.git`               | neither set                | `parent(git_common_dir)`         |

Submodules need `core.worktree` because their data lives in the parent's `.git/modules/`, so the implicit `parent(.git)` rule normal repos rely on would point at the wrong directory. Reading `core.worktree` from the bulk map is git's own signal for this — no `rev-parse` subprocess required.

## Correctness

The #762 → #1070 → #1280 chain of submodule fixes established two invariants: `repo_path()` must return the submodule working directory (not `.git/modules/sub`), and `list_worktrees()` depends on that to correct git's own mis-report for submodule main worktrees. Both still hold:

- `test_repo_path_in_submodule` (integration) passes — exercises `repo_path()`, `list_worktrees()`, and `worktree_for_branch()` in a real parent/submodule setup.
- `wt list` from inside a real submodule at `/tmp/subm-test/parent/sub` reports the submodule working directory, not the git data path.
- All 4 submodule-adjacent tests (`test_repo_path_in_submodule`, `test_push_{no_ff_,}with_submodule_recurse_config`, `test_remove_foreground_with_submodules`) pass.
- Bare repos still hit the `is_bare()` short-circuit; linked worktrees still fall through to `parent()` because the shared `.git` common dir has no `core.worktree`.

Errors still propagate via `Result` (from `all_config()` instead of the removed subprocess), preserving the #1280 invariant.

## Performance

hyperfine A/B on a healthy macOS system (200 runs each, `-N` bypasses shell):

| Context                  | Before          | After           | Speedup |
|--------------------------|-----------------|-----------------|---------|
| `wt noop` in normal repo | 53.2 ± 5.3 ms   | 48.8 ± 6.1 ms   | 1.09×   |
| `wt noop` in submodule   | 62.1 ± 12.6 ms  | 55.5 ± 11.7 ms  | 1.12×   |

Trace confirmation: `git rev-parse --show-toplevel [.git]` no longer appears in `RUST_LOG=debug` output for alias dispatch. The unrelated `rev-parse --show-toplevel [parent]` from `WorkingTree::root()` (legitimately successful) remains and is out of scope.

`benches/alias` results were too noisy to cite confidently on my machine during this session (`wt_version` — which this change doesn't touch — varied ±60% between runs), but the hyperfine measurement is clean.

Ref #2322

### DIFF
--- a/src/git/repository/mod.rs
+++ b/src/git/repository/mod.rs
@@ -80,6 +80,7 @@ use wait_timeout::ChildExt;
 use anyhow::{Context, bail};
 
 use dunce::canonicalize;
+use normalize_path::NormalizePath;
 
 use crate::config::{LoadError, ProjectConfig, ResolvedConfig, UserConfig};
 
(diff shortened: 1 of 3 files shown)


max-sixty:
Addressed in b4cefb254 — pre-populates the bulk config cache with an absolute `core.worktree` value to cover the branch, matching the pattern used by the existing `parse_config_list_z_*` unit tests. No filesystem fixture needed.

PR #2353, opened 2026-04-21
perf(alias): skip unused template-var accessors in hook context

max-sixty:
`build_hook_context` used to run `git rev-parse <branch>` for `{{ commit }}` and `git for-each-ref` for `{{ upstream }}` on every alias and hook dispatch, regardless of whether the pipeline referenced those vars. The `referenced_vars_for_config` set (already computed for `--KEY=VALUE` routing) now gates each accessor: a var is populated only if some template in the pipeline references it.

Addresses the parent-side piece of #2322: on macOS hosts with slow `execve` (~200 ms/fork), simple aliases now skip two guaranteed forks. On healthy systems each saved fork is ~3–5 ms, so the proportional win is similar but the absolute number is smaller.

## Scope

- `run_alias` and `prepare_steps` pass the `referenced_vars_for_config` union to gate accessors.
- Preview/debug paths (`wt step eval`, `wt switch --execute`, `wt hook <type>` preview, `wt step for-each`) pass `None` — they want the full context for inspection.
- Verbose mode (`-v`) at alias and hook dispatch sites passes `None`: the template-variable table prints `(unset)` for anything not populated, and users running `-v` are debugging and expect every var.

## Accurate error hints

`expand_shell_template` now takes an optional `ValidationScope`. When template expansion fails with an undefined-variable error, the "Available variables" hint lists every var the scope supports (via `vars_available_in`) rather than only the keys currently populated at runtime. Without this, lazy gating would narrow the hint — a template with a typo or a legitimately-unset var would see a shorter list omitting vars the pipeline didn't reference. The hint is meant to answer "what could you have typed," which is a property of the scope, not of this invocation's gated context.

As a bonus the hint now correctly lists scope-valid vars that weren't populated eagerly either (`upstream`, `owner`, `pr_number`, `pr_url`, `hook_name` for a pre-start hook), so users chasing "undefined value" for those vars see them suggested.

## Benchmarks (`benches/alias`, warm OS cache, macOS)

The bench matrix now covers three alias variants against the same repo: `noop` (zero vars), `commit` (`{{ commit }}`), `everything` (all gated vars via `| default`).

| variant | warm/1 | warm/100 | delta vs pre-#2340 noop baseline |
|---|---|---|---|
| noop | 45 ms | 45 ms | −18% / −22% |
| commit | 52 ms | 49 ms | noop + rev-parse fork |
| everything | 54 ms | 49 ms | noop + rev-parse + for-each-ref |

`commit`/`everything` land close to the pre-change baseline because they exercise the same accessors as before. The gap between `noop` and `commit`/`everything` (~4–7 ms) is the cost of the 2 forks lazy gating avoids — extrapolated to Mark's macOS (#2322) that's ~400 ms saved per simple-alias dispatch.

## Test plan

- [x] `cargo run -- hook pre-merge --yes` — 3318 tests, 0 failures; lints green
- [x] `RUST_LOG=worktrunk=debug wt noop` trace confirms noop alias drops from 7 → 5 git subprocesses in the parent; `wt showall` (references all gated vars) keeps 7

Co-Authored-By: Claude <noreply@anthropic.com>


max-sixty:
one downside of this is that it breaks the json piping to stdin

so I think we should leave to see how much it improves things once everything else merges

max-sixty:
Closing — main has since merged a stack of dispatch optimizations (`prewarm_info` batched rev-parse #2367, `head_sha()` cache reuse #2374, branch-inventory consolidation #2368/#2371) that already eliminate the rev-parse fork for `{{ commit }}` on the on-branch hot path. Merging main into this branch made the lazy-gating delta marginal on the typical alias path, so the remaining savings (cross-worktree `commit`, conditionally-forking `default_branch`/`primary_worktree`) don't justify the API expansion (`Option<&BTreeSet<String>>` threading, the verbose-mode bypass, and the bot-flagged JSON-contract narrowing on hook stdin).

Worth reassessing if a future repro shows alias dispatch is still slow on the affected macOS hosts after the recent batch of perf work lands in a release.

PR #2387, opened 2026-04-22
perf(ci-status): resolve remote-ref check via cached inventory

max-sixty:
`wt config state ci-status` spawned `git show-ref --verify --quiet refs/remotes/<branch>` on every invocation to decide whether the argument was a remote ref. Swap it for `Repository::is_remote_tracking_branch`, which reads the already-cached `remote_branches()` inventory populated by one `for-each-ref` scan. Same cleanup #2377 did for the sole other caller.

While there, collapse the surrounding `rev-parse` → `.unwrap_or_default()` → `if head.is_empty()` dance into a direct `.map_err(|_| BranchNotFound { ... })? .trim().to_string()` chain. The prior shape routed every git error through an empty-string sentinel; the new shape lets the error flow through directly.

### DIFF
--- a/src/commands/config/state.rs
+++ b/src/commands/config/state.rs
@@ -660,24 +660,17 @@ pub fn handle_state_get(
                 None => repo.require_current_branch("get ci-status for current branch")?,
             };
 
-            // Determine if this is a remote ref by checking git refs directly.
-            // This is authoritative - we check actual refs, not guessing from name.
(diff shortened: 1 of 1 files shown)


max-sixty:
Superseded by #2388, which reworked the same code block with a stronger fix (single `for-each-ref` query that also resolves the local/remote shadowing bug on branches literally named `origin/foo`). After merging main in, this branch's net diff against main is empty.

PR #2413, opened 2026-04-24
feat(step-copy-ignored): add TTY progress spinner

### LINKED ISSUE #2401: `wt step copy-ignored` skips re-ignored files after negation in `.gitignore`
### Description

When `.gitignore` contains a broad negation rule (e.g. `!**/src/main/**`) followed by a specific
re-ignore rule for a file (e.g. `src/main/resources/logback-spring.xml`), `wt step copy-ignored`
does not copy that file from the primary worktree — even though it is listed in `.worktreeinclude`
and git itself correctly considers it ignored.

---

### Environment

- **Application:** Spring Boot 4
- **Note:** The Spring Boot setup is not related to this issue; the bug is purely in `wt`'s gitignore parsing logic.

---

### Steps to Reproduce

1. Add the following to `.gitignore` (the negation rule is generated automatically by Spring Boot Initializr):
   ```
   !**/src/main/**
   src/main/resources/logback-spring.xml
   ```

2. Add `src/main/resources/logback-spring.xml` to `.worktreeinclude`.

3. Configure `wt.toml` to run `wt step copy-ignored` as a post-start step:
   ```toml
   [post-start]
   copy-include = "cp {{ primary_worktree_path }}/.worktreeinclude {{ worktree_path }}/"
   copy = "wt step copy-ignored"
   ```

4. Run `wt switch -c new-feature`.

---

### Expected Behavior

`src/main/resources/logback-spring.xml` is copied from the primary worktree into the new worktree,
consistent with all other entries in `.worktreeinclude`.

### Actual Behavior

`src/main/resources/logback-spring.xml` is **not** copied. All other entries in `.worktreeinclude`
(`.env`, `.venv/`, `build/`) are copied successfully.

---

### Root Cause Hypothesis

Git resolves ignore rules by **last-match-wins**: the negation `!**/src/main/**` un-ignores the
file, but the subsequent specific rule `src/main/resources/logback-spring.xml` re-ignores it.
Running `git check-ignore -v src/main/resources/logback-spring.xml` confirms that git correctly
treats the file as ignored.

`wt step copy-ignored` appears to use its own gitignore parsing logic that does not correctly handle
this two-step pattern, causing it to treat the file as *not* ignored and therefore skip it.

---

### Suggested Fix

Use `git check-ignore <path>` (or the equivalent libgit2 API) as the authoritative source for
determining whether a file is ignored, rather than a custom gitignore parser. This ensures parity
with git's own last-match-wins resolution.

---

### Workaround

Remove the broad negation patterns from `.gitignore` so the specific re-ignore rule is unambiguous:

```diff
- !**/src/main/**
- !**/src/test/**
```

max-sixty:
Large copies of `node_modules`, `build/`, or `target/` ran silently — if the walk or the copy was slow, users couldn't tell whether `wt step copy-ignored` was making progress or hung. Adds a stderr-only spinner: `⠼ Copying 7,272 files · 64.5 MiB`, updated in place at ~10Hz, only while a real TTY is on the other end. The post-copy summary repeats the same units (`✓ Copied 7,272 files · 64.5 MiB`) so the spinner's information isn't dropped on completion.

Gate: stderr-is-TTY, `verbosity == 0` (verbose mode already lists every entry), and not `--dry-run`. A 300ms startup delay keeps sub-second copies silent — short copies never render a line at all. Non-TTY output is byte-for-byte unchanged apart from the new byte total in the summary.

### Approach

- New `src/copy_progress.rs`: `CopyProgress` is `Option<Inner>` — disabled is `None`, enabled holds an `Arc<Shared>` of atomic counters plus a crossterm ticker `JoinHandle`. `start()` is named deliberately (not `new()`) because it spawns a thread — `Default` semantics would be misleading. `Drop` joins the ticker and clears the line.
- `copy_leaf` returns `Result<Option<u64>>` (bytes on copy, `None` on skip). Bytes come from the `symlink_metadata` we already stat, so no extra syscall — actually one fewer on Unix since we now reuse it for permissions too.
- `copy_dir_recursive` returns `Result<(usize, u64)>` and takes `&CopyProgress`; rayon workers bump `progress.file_copied(bytes)`. The summary in `step_copy_ignored` reuses the totals.
- New skill section in `.claude/skills/writing-user-outputs/SKILL.md` pointing future authors at `CopyProgress` as the spinner pattern.

### Testing

Unit tests cover `format_count`, `format_bytes` (IEC), `format_line`, `render_line`/`clear_line` (against an in-memory `Vec<u8>`), the disabled no-op path, and the enabled ticker lifecycle (via a test-module-local helper that bypasses the TTY check). Manual smoke via `script(1)`: big copies render the in-place spinner cleanly; small copies (<300ms) render nothing beyond the summary; pipes, `-v`, and `--dry-run` are unchanged.

`codecov/patch` covers ~94% of the diff. The remaining ~10 uncovered lines are the enabled branch of `start()` (Arc allocation + thread spawn), which can't be reached without a real TTY in tests. Equivalent logic is exercised by the test-module helper.

Closes #2401. Thanks to @tehdb for the suggestion.


tehdb:
@max-sixty Wow, that was incredibly fast, thank you! The spinner is a great addition and really appreciate the attention to detail with the 300ms startup delay so short copies stay clean. 🙏

PR #2488, opened 2026-04-30
refactor(hooks): retire Option<&mut HookAnnouncer> plumbing

max-sixty:
## Summary

Now that all callers (merge, remove single + multi, prune, step commit) construct an announcer, the receiving sites take it by mutable reference rather than `Option`. `handle_remove_output` and the four internal handlers lose the `Option` wrapper, and `spawn_hooks_after_remove` collapses to a single `announcer.extend(...)` call. `CommitOptions::commit` likewise; the standalone `step_commit` path now constructs a local announcer and flushes it after `commit()` returns.

`RemovedWorktreeOutputContext` no longer carries `show_branch_in_hooks` — the announcer already knows — and `handle_remove_output` drops the matching parameter.

`handle_squash` keeps `Option<&mut HookAnnouncer<'_>>`: it loads its own config inside, so the standalone path builds a local announcer from that owned config rather than relying on a caller-owned one.

## Test plan
- [x] `pre-commit run --all-files`
- [x] `cargo test --test integration --features shell-integration-tests` (1725 passed)

### DIFF
--- a/src/commands/commit.rs
+++ b/src/commands/commit.rs
@@ -8,9 +8,7 @@ use worktrunk::styling::{
 
 use super::command_executor::CommandContext;
 use super::command_executor::FailureStrategy;
-use super::hooks::{
-    HookAnnouncer, execute_hook, prepare_background_pipelines, run_hooks_background,
-};
+use super::hooks::{HookAnnouncer, execute_hook, prepare_background_pipelines};
 use super::repository_ext::warn_about_untracked_files;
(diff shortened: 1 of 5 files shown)


max-sixty:
Superseded by #2484, which landed a strict superset of these simplifications (also covers handle_squash, hook_commands::run_post_hook, picker post-remove, and makes run_hooks_background private). Should have noticed before pushing — that dispatched session was "hooks-dedup".

PR #2502, opened 2026-04-30
Scrub legacy directive file from child commands

douglas:
## Summary

Fixes legacy directive-file passthrough into child commands.

The split directive protocol already prevents alias and hook child processes
from writing arbitrary shell into the parent session: children may receive
`WORKTRUNK_DIRECTIVE_CD_FILE`, but not `WORKTRUNK_DIRECTIVE_EXEC_FILE`.
However, the compatibility variable `WORKTRUNK_DIRECTIVE_FILE` still combined
CD and shell execution directives for old shell wrappers. Passing that legacy
file through to aliases, foreground hooks, or concurrent child commands let
those children append shell that the parent wrapper could later source.

This PR keeps top-level legacy wrapper compatibility intact, but stops passing
`WORKTRUNK_DIRECTIVE_FILE` into child command execution. Child aliases, hooks,
and concurrent commands now only receive the safe CD directive file.

## Changes

- Remove legacy directive passthrough from `DirectivePassthrough`.
- Remove `shell_exec::Cmd::directive_legacy_file`.
- Stop concurrent command spawning from re-adding `WORKTRUNK_DIRECTIVE_FILE`.
- Update comments documenting the new trust boundary.
- Add regression coverage for aliases and foreground hooks under legacy env
  setup, plus unit coverage for legacy env filtering.

## Tests

- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --bin wt test_inherit_from_env_ignores_legacy_directive_file --features shell-integration-tests`
- `cargo test --test integration legacy_directive_file --features shell-integration-tests`

### DIFF
--- a/src/commands/alias.rs
+++ b/src/commands/alias.rs
@@ -28,7 +28,9 @@
 //! first, then project. The CD directive file is passed through to child
 //! processes so inner `wt` invocations can redirect the parent shell's cwd;
 //! the EXEC directive file is scrubbed so alias bodies cannot inject
-//! arbitrary shell into the interactive session.
+//! arbitrary shell into the interactive session. The legacy single-file
+//! directive is handled only by top-level legacy output mode, not by alias
+//! child execution.
 
 use std::collections::{BTreeMap, BTreeSet};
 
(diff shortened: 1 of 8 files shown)


max-sixty:
@worktrunk-bot we can probably remove the old directive form completely; do you want to make that change here?

max-sixty:
yes, OK, do a new PR please

PR #2510, opened 2026-04-30
ci: harden GitHub Actions workflows

douglas:
## Problem We Are Solving

The GitHub Actions workflows had broader default token permissions than many jobs need. Broad permissions increase the blast radius if a job, action, or workflow step is compromised.

The riskiest workflow shape is `pull_request_target` combined with privileged tokens and PR-related checkout logic. That event can access secrets and write-scoped tokens, so it must not check out or execute untrusted fork PR code.

The repository also did not have a dedicated CodeQL workflow for Rust and GitHub Actions security patterns, so risky workflow changes depended primarily on manual review.

## Why This Matters

CI is part of the developer-machine safety boundary for this project. If a privileged workflow executes untrusted code, an attacker can potentially exfiltrate tokens, modify PRs, or affect the repository. Least-privilege permissions and explicit fork quarantine reduce that risk without disabling useful automation for trusted same-repo PRs.

## What Changed

- Added restrictive workflow-level permissions across GitHub Actions workflows.
- Kept write permissions only on jobs that need them.
- Hardened `tend-review` so fork PRs do not run the privileged checkout/setup/Tend execution path.
- Added a fork notice step so skipped fork reviews are visible and intentional.
- Added a CodeQL workflow for Rust and GitHub Actions with `security-extended` queries.

## How This Solves It

The permission changes make read-only the default and force write scopes to be explicit per job. That narrows the token permissions available to unrelated workflow steps.

The `tend-review` quarantine keeps same-repo automation working, but fork PRs now stop before the privileged workflow checks out or executes PR code. This preserves the useful bot review path where it is lower risk and avoids the dangerous `pull_request_target` plus untrusted-code combination for forks.

CodeQL adds automated scanning for Rust code and GitHub Actions workflow security issues, giving maintainers another line of review for future workflow changes.

## Validation

- YAML sanity check passed with `ruby -rpsych` over all workflow files.
- `git diff --check` passed.
- `pre-commit run --all-files` could not run because `pre-commit` is not installed locally.
- `actionlint` is not installed locally.

## Reviewer Notes

- Same-repo Tend review automation remains enabled.
- Fork PRs now receive a notice and skip privileged checkout/setup/Tend execution.
- The generated Tend workflow headers are preserved; this PR makes a minimal security hardening edit directly in the generated workflow files.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
   # CI does clean builds, so incremental compilation tracking adds I/O overhead
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,40 @@
+name: codeql
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '23 3 * * 1'
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [actions, rust]
+    steps:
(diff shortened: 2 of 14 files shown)


max-sixty:
@douglas appreciate the efforts, but these need much more of your gating; this one would not work at all with `tend`, for example

douglas:
> @douglas appreciate the efforts, but these need much more of your gating; this one would not work at all with `tend`, for example

Got it! I'll review this implementation in the future as I get more acquainted with the project :+1:

PR #2511, opened 2026-04-30
test: add parser and path fuzzing harnesses

douglas:
## Problem We Are Solving

Worktrunk has several developer-safety-sensitive parser and path surfaces:

- git remote URL parsing
- config migration
- template syntax validation and variable reference detection
- filename sanitization

Unit tests cover known cases, but they do not continuously explore malformed, adversarial, or unusual inputs. Bugs in these areas can lead to crashes, incorrect project identity, unsafe filenames, or bad config migrations.

## Why This Matters

This project runs on developer machines and manipulates repositories, config, paths, hooks, and command execution state. Parser and path-boundary mistakes are a practical way for malformed repository state or config to become a developer-machine integrity issue.

Fuzzing gives maintainers a repeatable way to stress these surfaces beyond hand-written examples.

## What Changed

- Added a standalone `fuzz/` package for `cargo-fuzz`.
- Added a fuzz target for `GitRemoteUrl::parse`.
- Added a fuzz target for `worktrunk::config::migrate_content`.
- Added a fuzz target for `validate_template_syntax` and `template_references_var`.
- Added a fuzz target for `sanitize_for_filename`.
- Added `fuzz/README.md` with install, build, and run commands.
- Kept the fuzz package separate from normal stable builds so regular `cargo test` and pre-merge do not require nightly.

## How This Solves It

The fuzz targets exercise public parser/sanitizer APIs with arbitrary input and assert basic invariants:

- parsing should not panic
- migrations should be stable enough to run on arbitrary text
- template validation/reference detection should tolerate malformed templates
- filename sanitization should produce usable filenames without path separators

The harnesses are intentionally lightweight. They provide the scaffolding needed for local or scheduled fuzzing without adding nightly-only requirements to normal CI.

## Validation

- `cargo metadata --manifest-path fuzz/Cargo.toml --no-deps` passed.
- `cargo check --manifest-path fuzz/Cargo.toml --bins` passed.
- `cargo fmt --manifest-path fuzz/Cargo.toml --all --check` passed.
- `cargo +nightly fuzz build` could not run because `cargo-fuzz` is not installed locally.
- `cargo test --locked --lib --bins` currently fails on the existing unrelated `testing::mock_commands::tests::test_mock_config_write` test.

## Reviewer Notes

- The path-boundary fuzz target is documented as pending because the relevant copy-boundary helper is still private on this branch base.
- No Miri gate is added here. This PR only adds fuzzing scaffolding and leaves nightly-only or scheduled fuzz execution policy for a later decision.


max-sixty:
- if someone can find a problem using a fuzzer, then we'll fix the problem
- if folks can find lots of problems using fuzzers, then we could add a fuzzer
- but until then, a lot of this is AI cruft, respectfully

PR #2598, opened 2026-05-04
refactor(list): always emit main/remote JSON fields

max-sixty:
## Summary

`wt list --format=json` now always emits the `main` and `remote` objects;
inner values are explicit `null` when the data is not applicable or
has not been computed yet.

Before:
```json
// main worktree
{ "branch": "main", "main_state": "is_main", ... }
// branch with no upstream
{ "branch": "feature-a", "main": {"ahead": 1, "behind": 0}, ... }
```

After:
```json
// main worktree
{ "branch": "main", "main_state": "is_main",
  "main":   { "ahead": null, "behind": null, "diff": null },
  "remote": { "name": "origin", "branch": "main", "ahead": 0, "behind": 0 }, ... }
// branch with no upstream
{ "branch": "feature-a",
  "main":   { "ahead": 1, "behind": 0, "diff": null },
  "remote": { "name": null, "branch": null, "ahead": null, "behind": null }, ... }
```

## Why

This is a JSON output shape change (new behavior: explicit nulls
instead of omitted fields). The old shape required consumers to handle
both "field absent" and "field present-but-zero", because `main` was
absent for the main worktree and `remote` was absent for branches
without a tracking branch. The new shape gives consumers a single,
stable schema.

CLAUDE.md's "External interfaces to protect" lists config and CLI
flags but not JSON output, so this surface is flexible. The change is
also forward-compatible: existing jq queries like `.remote.ahead > 0`
keep working because `null > 0` is falsy.

## Notes on the model layer

The TODOs at `src/commands/list/model/item.rs:208-250` referenced JSON
shape, but the user-facing JSON travels through `JsonItem` in
`src/commands/list/json_output.rs` (constructed via
`JsonItem::from_list_item`), not through the flatten serialization on
`ListItem`. The fix lives at that real serialization boundary; the
model-side comments now point at it.

## Test plan

- [x] `cargo run -- hook pre-merge --yes` — 3484 tests, 0 failures
- [x] Manual: `wt list --format=json` shows new shape on a real repo
- [x] Snapshots regenerated (5 list_json + 2 help snapshots)
- [x] Docs synced (`docs/content/list.md`, `skills/worktrunk/reference/list.md`)

### DIFF
--- a/docs/content/list.md
+++ b/docs/content/list.md
@@ -172,8 +172,8 @@ Query structured data with `--format=json`:
 | `main_state` | string | Relation to the default branch (see below) |
 | `integration_reason` | string | Why branch is integrated (see below) |
 | `operation_state` | string | `"conflicts"`, `"rebase"`, or `"merge"`; absent when clean |
-| `main` | object | Relationship to the default branch (see below); absent when is_main |
-| `remote` | object | Tracking branch info (see below); absent when no tracking |
+| `main` | object | Relationship to the default branch (see below); always present (inner fields are `null` for the main worktree) |
+| `remote` | object | Tracking branch info (see below); always present (inner fields are `null` when no tracking branch) |
 | `worktree` | object | Worktree metadata (see below) |
 | `is_main` | boolean | Is the main worktree |
 | `is_current` | boolean | Is the current worktree |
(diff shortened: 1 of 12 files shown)


max-sixty:
Closing without merging.

On reread, this is a lateral move at best, and arguably worse:

- Creates inconsistency within the same struct — `worktree`, `commit`, `operation_state`, `diff` still use `skip_serializing_if`, while `main` / `remote` are always present. Worse than either pure rule.
- Redundant signaling — `main_state: "is_main"` already conveys "this is the main worktree." Adding `main: { ahead: null, behind: null }` doesn't add information.
- Output bloat — every row now carries 3–7 placeholder null fields. Noticeable for `wt list --json | jq` pipelines.
- No actual consumer was broken. The original TODO was a "consumers *may* expect this" hypothesis with no filed report. Klotz-test fails: we'd be adding structure for a non-problem.

The "absent when not applicable" convention is more self-documenting and lighter. Will follow up with a small PR removing the now-obsolete TODOs in `item.rs` so the question doesn't get re-litigated.

PR #2602, opened 2026-05-04
perf(list): skip redundant MergeTreeConflicts on dirty worktrees

max-sixty:
## Summary

- For dirty worktrees, `WorkingTreeConflicts` already runs `git merge-tree` against a tree that includes HEAD + uncommitted changes; `MergeTreeConflicts` was duplicating that work against HEAD alone.
- `MergeTreeConflictsTask` now peeks the shared porcelain cache and short-circuits with a sentinel `false` when the worktree is dirty without unmerged entries.
- `tier_would_conflict` now treats `Some(Some(_))` as authoritative — the dirty-tree probe wins regardless of the HEAD probe. This is the answer most users want for "would this merge conflict?" anyway, since the dirty tree reflects the current working state.

## Implementation notes

- The optimization fires only for worktree items that are dirty AND have no unmerged entries. Branch-only items, clean worktrees, and unmerged-state worktrees still run the HEAD probe normally.
- One snapshot updated: `test_list_warns_when_commit_details_batch_fails` — the corrupt-HEAD repro now surfaces `merge-tree-conflicts (fatal: bad object HEAD)` from the early porcelain check instead of the deeper `merge-base failed` message. Same severity; same diagnostic value.
- Behavior change in one edge case: when HEAD would conflict but uncommitted changes resolve it, the dirty probe now wins (rules out the conflict). The TODO this PR resolves explicitly proposed this trade-off.

## Test plan

- [x] `cargo run -- hook pre-merge --yes` (full lints + 3484 tests) passes
- [x] Manually verified in a synthetic dirty-conflict repo via `RUST_LOG=worktrunk=debug` that exactly one merge-tree subprocess fires per dirty row (the `WorkingTreeConflicts` ephemeral-commit one), down from two
- [x] `test_list_working_tree_conflicts` and `test_list_full_clean_working_tree_uses_commit_conflicts` both pass — `✗` symbol still renders correctly for dirty conflicts; clean conflicts still detected via HEAD probe
- [x] Updated `tier_would_conflict` unit tests cover the new short-circuit and the "skipped HEAD probe" path

### DIFF
--- a/src/commands/list/collect/execution.rs
+++ b/src/commands/list/collect/execution.rs
@@ -375,11 +375,9 @@ pub fn work_items_for_worktree(
         TaskKind::UserMarker,
         TaskKind::WorkingTreeConflicts,
         TaskKind::BranchDiff,
-        // TODO: For dirty worktrees, WorkingTreeConflicts already runs merge-tree
-        // (via stash-create + merge-tree). MergeTreeConflicts duplicates that call
-        // against HEAD. Could skip MergeTreeConflicts when WorkingTreeConflicts
-        // produces a non-None answer, but needs result-ordering changes since both
-        // tasks run in parallel today.
+        // MergeTreeConflictsTask peeks the shared porcelain cache and
+        // skips its own merge-tree call when WorkingTreeConflictsTask
+        // will produce an authoritative dirty-tree result.
         TaskKind::MergeTreeConflicts,
         TaskKind::CiStatus,
         TaskKind::WouldMergeAdd,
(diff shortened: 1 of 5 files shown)


max-sixty:
Pushed `cb507e25c`:

- Refactored `tier_would_conflict` to a single tuple match — exhaustive without `unreachable!()`, closes the codecov/patch gap.
- Added the `(Some(true), Some(Some(false)))` test pinning WT-overrides-positive-HEAD as the strongest demonstration of the new authoritative-WT contract.

Walked all ten input shapes against both implementations to confirm semantic equivalence. `wt hook pre-merge --yes` green locally (3484/3484, no snapshot drift).

PR #2603, opened 2026-05-04
feat(hints): track shown count and escalate shell-integration hint at 5+

max-sixty:
## Summary

- Migrate `worktrunk.hints.<name>` from `"true"` to an integer counter.
  `mark_hint_shown` now reads the existing value, increments, and writes
  back; new `hint_count` exposes the total.
- Use the count to escalate the shell-integration install hint — after
  5+ prior displays, append a `wt config show` pointer so users who keep
  seeing it can investigate why their wrapper isn't intercepting.
- Legacy `"true"` values parse as 0 (documented), so the next mark
  normalises to 1. `has_shown_hint` still reports presence, so first-time
  branches still skip for legacy users.

Resolves the TODO at `src/output/shell_integration.rs` that flagged the
infrastructure change as a prerequisite.

## Test plan

- [x] New unit tests in `src/git/repository/config.rs` cover increment,
      legacy `"true"` migration, and clear-resets-counter.
- [x] Updated `shell_integration_hint` snapshot covers count 0/4 (base)
      and 5/10 (escalated).
- [x] `cargo run -- hook pre-merge --yes` (3485 tests, all lints) green.


max-sixty:
Pushed `e20476382` updating the `wt config state hints` help text to describe the new integer-count storage. `wt hook pre-merge --yes` green locally (3485/3485, no snapshot drift). The macOS spawn flake is, per your read, unrelated; the rerun should clear it.

src/cli/config.rs
```diff
@@ -782,7 +782,7 @@ $ wt config state logs clear
     /// One-time hints shown in this repo
     #[command(
         after_long_help = r#"Some hints show once per repo on first use, then are recorded in git config
-as `worktrunk.hints.<name> = true`.
+as `worktrunk.hints.<name>`, an integer count of times the hint has been shown.
```
max-sixty:
```suggestion
as `worktrunk.hints.<name>`, a count of times the hint has been shown.
```

?

PR #2641, opened 2026-05-07
feat(config): add codename template filter

endigma:
- Adds a `codename(n)` filter for generating friendly names deterministically
- Has a few tests, and documents possible issues with collisions, along with recommendations to avoid them


max-sixty:
great idea @endigma ! thank you!

couple of questions:
- when do you see using this vs the branch name itself? I could see it being helpful if someone thought "I don't know what I'm going to do so I just want a name". but at the moment I think we still need a name?
- rather than suggesting to "pair with hash", should we just massively increase the number of adjectives & nouns, so it's difficult to get a conflict? IIUC `codename(3)` with a big enough sample is all atoms in the solar system or similar... (@worktrunk-bot is there a rust dependency that has lots of words? otherwise we could just write them out ourselves)

max-sixty:
> One small thing worth keeping regardless of list size: the 6-word adjective/noun overlap I flagged earlier (`cedar`, `dawn`, `hazel`, `hollow`, `maple`, `marble`) — at any list size, it's worth either stripping the overlaps from one side or detecting `adjective == noun` in `codename()` to avoid `maple-maple` style outputs.

(this is fine as-is, fwiw)

I'll let @endigma decide on the questions above

endigma:
@max-sixty its so the directory name isn't the branch name, which can be really annoying and long for some branch naming conventions (e.g. ticket number + full sanitized title) and also my prompt already writes the branch name

max-sixty:
> @max-sixty its so the directory name isn't the branch name, which can be really annoying and long for some branch naming conventions (e.g. ticket number + full sanitized title) and also my prompt already writes the branch name

ah that makes sense. that's a clever approach

---

I think let's try using `petname` and make a note to avoid upgrading if the words change. we can always vendor it if they decide to change their words. then we can drop the hash

@worktrunk-bot could you make this change, and @endigma can say if they would prefer not to

endigma:
Bigger wordlist is good in my book, I was probably going to run codename(2) with no hash anyway lol

In the unlikely event of a collision, does everything error out clean so the user could just pick an ever so slightly different branch name? 

@worktrunk-bot can you answer this?

endigma:
Cool bot, does this need anything else to be merged @max-sixty ?

PR #2715, opened 2026-05-11
Support Omarchy dynamic themes in switch picker

douglas:
## Summary

- read Omarchy's active `colors.toml` when the `wt switch` picker opens
- map Omarchy theme colors into Skim's `--color` scheme when that Omarchy theme file is present and valid
- preserve the existing built-in picker palette for non-Omarchy users, missing theme files, and invalid color data
- document the behavior in switch help, generated docs, skill reference, and the help snapshot
- address draft review feedback by using the project home-directory helper and directly testing the fallback resolver path

### Omarchy / Catpuccin

screenshot-2026-05-11_15-25-09

### Omarchy / Evergreen

screenshot-2026-05-11_15-24-31

## User Impact

This only changes the `wt switch` interactive picker when Worktrunk is running in an Omarchy environment with a valid active theme at `~/.config/omarchy/current/theme/colors.toml`.

For users not on Omarchy Linux, nothing changes: the picker continues to use Worktrunk's existing built-in color palette. Omarchy users also get the same fallback behavior if the theme file is missing, unreadable, or invalid.

Theme changes are picked up the next time the picker opens. An already-running picker is not repainted after an Omarchy theme switch because Skim receives its color scheme at startup.

## Commit Structure

1. `Support Omarchy colors in switch picker`
   - Adds the theme resolver/parser and switches the picker from a fixed color string to dynamic resolution.
2. `Document Omarchy switch picker theming`
   - Documents that the picker reads Omarchy colors each time it opens and that theme changes apply on the next picker invocation.
3. `Address Omarchy picker review feedback`
   - Uses Worktrunk's `home_dir` re-export and makes the fallback color resolver directly testable.

## Validation

- `cargo test picker::theme`
- `cargo test --test integration switch_picker_dry_run`
- `cargo test --test integration test_docs_are_in_sync`
- `cargo insta test --accept -- --test integration "test_help"`
- `cargo test --test integration test_help`
- `cargo clippy --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/docs/content/switch.md
+++ b/docs/content/switch.md
@@ -58,6 +58,8 @@ Shortcuts also apply to `--base`. For a fork PR/MR, the head commit is fetched a
 
 When called without arguments, `wt switch` opens an interactive picker to browse and select worktrees with live preview.
 
+On Omarchy, the picker reads `~/.config/omarchy/current/theme/colors.toml` each time it opens and uses the active theme's colors. Theme changes apply the next time the picker opens. When that file is unavailable, the picker falls back to Worktrunk's built-in palette.
+
 <figure class="demo">
 <picture>
   <source srcset="/assets/docs/dark/wt-switch-picker.gif" media="(prefers-color-scheme: dark)">
--- a/skills/worktrunk/reference/switch.md
+++ b/skills/worktrunk/reference/switch.md
@@ -60,6 +60,8 @@ Shortcuts also apply to `--base`. For a fork PR/MR, the head commit is fetched a
 
 When called without arguments, `wt switch` opens an interactive picker to browse and select worktrees with live preview.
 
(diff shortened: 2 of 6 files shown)


max-sixty:
thanks for the PR

but I'm not really sure it's that useful; we're not doing that much theming

what's the case for deviating from terminal colors?

douglas:
> thanks for the PR
> 
> but I'm not really sure it's that useful; we're not doing that much theming
> 
> what's the case for deviating from terminal colors?

Actually, I'll close this - I just trying in another laptop and it works so maybe it was something on my setup that was broken.

Sorry for the noise, @max-sixty

max-sixty:
no prob!

PR #2718, opened 2026-05-11
perf(list): cold-start primer for the Remote⇅ column, per-upstream-group

max-sixty:
## Summary

- Mirrors the `main↕` cold-start primer from #2704 for the `Remote⇅` column. `Repository::prime_upstream_ahead_behind_cache` runs one `for-each-ref %(ahead-behind:UPSTREAM_SHA)` walk per unique upstream SHA (scoped to that upstream's branches) and bulk-writes the `ahead-behind/` cache. Per-row `UpstreamTask::compute` reads the cache via `ahead_behind_by_sha`.
- Nested inside the snapshot spawn in the post-skeleton `rayon::scope` so it reads the snapshot's already-scanned local/remote inventories — no extra `for-each-ref refs/remotes/` fork. Candidates scoped to branches that actually render an `UpstreamTask` row (worktree-attached when `!show_branches`). Primer honors `list.task-timeout-ms` via `set_command_timeout`.
- Cache-correctness defenses match #2704: `%(ahead-behind:UPSTREAM_SHA)` (not `:refname`) so git counts against the SHA the cache will be keyed by; cache key uses the `%(objectname)` git reports in the same walk; writes flow through `put_ahead_behind_bulk` (one `sweep_lru` at end); orphan branches normalize to `(0, 0)` via per-upstream memoized `rev-list --count`.

### Known limitation

When every branch tracks its own remote (distinct upstream SHA per branch), every group has `refs.len() == 1` and the primer skips all of them — those fall through to the per-row parallel path. A `%(upstream:track)` walk could batch them, but git computes that atom against the upstream's current value at walk time, which we can't pin to the SHA the cache will be keyed by; the resulting race would break the cache invariant `compute_ahead_behind` establishes on a miss. Documented in the primer's docstring.

### Before/after on a 10-branch shared-upstream fixture

Cold `wt list --branches` trace (`RUST_LOG=worktrunk=debug`):

- `for-each-ref --format=%(refname) %(objectname) %(ahead-behind:<main_sha>) refs/heads/` — snapshot's `main↕` walk (unscoped)
- `for-each-ref --format=%(refname) %(objectname) %(ahead-behind:<feat-base_sha>) refs/heads/feat-base refs/heads/work1 ... refs/heads/work10` — primer's scoped Remote⇅ walk
- 23 cache entries written (12 keyed by main, 11 keyed by feat-base)

Warm run: `grep ahead-behind:` returns 0 — both walks skipped, cache hits everywhere.

## Test plan

- [x] 9 new primer tests in `ref_snapshot.rs` mirror the `capture_ahead_behind` patterns (cold→warm tampered-cache survival, no-upstream / `[gone]` skips, orphan normalization, sub-threshold skip, two-upstream-group routing, equal-branch `(0,0)`, local-upstream `branch.X.remote = .` resolution)
- [x] `cargo run -- hook pre-merge --yes` — 3664 tests pass
- [x] `/review-codex` iterated to convergence on cache correctness, candidate scoping, redundant remote-scan, and timeout coverage


max-sixty:
Confirming both:

1. **codecov base-pointer**: yes, the `parent: bfb117a665...` in `commit_uploads` is stale — `git merge-base origin/main remote-ahead-behind-batch` returns `5384f62b`, so every line #2704 added is being charged to this PR's patch. Existing infrastructure issue, not a real coverage gap on this PR.

2. **line 426 (`(0, 0)` orphan arm)**: cargo-llvm-cov mismatched-data, same class as 412/415/437. Verified by running the path: with `main` at N commits and an orphan branch, `git for-each-ref --format='%(ahead-behind:main_sha)' refs/heads/orphanbr` returns `1 N` (ahead=1, behind=N), and `rev-list --count main_sha` returns N. Both equal → line 426 fires, producing `(0, 0)` — which is exactly what `prime_upstream_normalizes_orphan_to_zero` asserts. The else arm on 428 would have produced `(1, N)`, failing the assert.

Local `task coverage` reports `warning: 25 functions have mismatched data` and the same lines (412, 415, 426, 437) are missing in local cobertura too, matching #2704's CLAUDE.md note about [LLVM #97574](https://github.com/llvm/llvm-project/issues/97574).

PR #2751, opened 2026-05-13
feat(plugin): /wt-switch-create — optional repo arg and `--` task delimiter

max-sixty:
The `/wt-switch-create` skill now accepts `<branch> [<repo>] [-- <task>]`. The optional second token names a different repository to create the worktree in (instead of the session's current one); the skill `cd`s into that repo with a `Bash` call before invoking `EnterWorktree` (which has no repo parameter), then verifies the new worktree landed under the requested repo as a safety check.

The `--` cleanly separates the task from the rest. Without one, the parser treats a path-shaped second token (absolute, `~`-relative, `./`/`../`-relative, or an existing directory) as the repo and the remaining tokens as the task — anything else after the branch is task text, so the old shorthand `/wt-switch-create my-branch fix the bug` keeps working.

Examples in the skill:

```
/wt-switch-create my-feature -- fix the parser bug
/wt-switch-create my-feature ~/workspace/other-repo -- fix the parser bug
/wt-switch-create my-feature
```

Also tidied existing prose: the nesting + different-repo case now stops rather than silently running the task in the wrong repo, and the scope note mentions the repo arg.

The follow-up commit updates the public docs page (`docs/content/claude-code.md`) so the `/wt-switch-create` signature shown at worktrunk.dev matches the new grammar; `skills/worktrunk/reference/claude-code.md` re-synced via `test_docs_are_in_sync`.

No code or test changes.

### DIFF
--- a/docs/content/claude-code.md
+++ b/docs/content/claude-code.md
@@ -73,7 +73,7 @@ Claude Code agents can run in isolated worktrees (`isolation: "worktree"`). By d
 
 ## `/wt-switch-create` command
 
-`/wt-switch-create <branch> [task...]` starts work in a fresh worktree without leaving the session. It creates (or re-enters) the named worktrunk worktree — sibling layout `<repo>.<branch>/`, not `.claude/worktrees/` — switches the session's working directory into it, then runs the rest of the arguments as the task there. The command rides the same `WorktreeCreate` hook as agent isolation, so the worktree gets worktrunk's naming, hooks, and lifecycle.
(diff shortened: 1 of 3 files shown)


max-sixty:
Thanks — pushed [`1eec508`](https://github.com/max-sixty/worktrunk/commit/1eec50810) updating `docs/content/claude-code.md` (and the synced `skills/worktrunk/reference/claude-code.md`) to show the new `<branch> [<repo>] [-- <task>]` signature. PR description updated to mention the docs follow-up.

Issue #1888, opened 2026-04-02
`wt step prune` seems very slow


ortonomy:
<img width="854" height="188" alt="Image" src="https://github.com/user-attachments/assets/1c960a37-9139-4164-a257-7066d6902da7" />

willing to accept that if there are calls to github under the hood - it might be being affected by high latency to github from my network, but this call took > ~3 minutes to prune a single tree. 

This seems excessive.

ortonomy:
3 worktrees

<img width="745" height="141" alt="Image" src="https://github.com/user-attachments/assets/5af478d1-5cfe-4497-8360-6bf58c1bc2a1" />

```sh
wt list --format=json | jq length
4
``

4 if you include `master`

**Comment by ortonomy** [comment]
```sh
git branch | wc -l
     112
```

112!

ortonomy:
```sh
wt --version
wt 0.33.0
```

max-sixty:
@worktrunk-bot  good diagnosis

please could you draft a change to do this?

thanks for the issue @ortonomy

razor-x:
I noticed an odd behavior on projects with many branches with `wt step prune`. It happens consistently for me but it might be hard to reproduce in an isolated test. Might be related to the PR for this issue.

Basically, I'll run prune, it will speed though a bunch of branches, then stall out, then I'll kill it with ctrl-c, then I'll do it again, will do a few more branches, then stall out, etc, etc. It looks like this.

If I let it run it will keep going, but the UX makes it seem like it got stuck.

<img width="819" height="772" alt="Image" src="https://github.com/user-attachments/assets/7f491d4b-5b8c-48ec-a8a0-d0b9c73feded" />

max-sixty:
> A slow `pre-remove` hook in your config would look exactly like "speed through, then stall", with Ctrl-C dropping the in-flight hook so the next run picks up the rest.

do we show the pre-remove hook outputs when we're running `wt step prune`?

PR #2752, opened 2026-05-13
perf(integration): bound patch-id squash-merge scan to 500 commits

max-sixty:
## Summary

`is_squash_merged_via_patch_id` runs `git log -p {merge-base}..{target} | git patch-id` over the *entire* target-side history. The range is unbounded: on a fast-moving repo with an old branch — tens of thousands of commits since divergence — a single integration check takes seconds to tens of seconds, which surfaces as `wt step prune` (and `wt list`) going visibly silent on the parallel check tail in #1888.

This caps the scan at `PATCH_ID_SCAN_MAX_COMMITS = 500` commits via a cheap graph-only `git rev-list --count` pre-flight. Above the cap, the check returns `Ok(false)` — the safe direction (branch kept, not wrongly deleted); `wt remove -D` still removes branches that fall past the cap.

## Why 500

Count is a rough proxy — per-commit cost scales with `changed_files × changed_lines`, not just count. Working back from "keep one check under a few seconds":

- typical repo (~5-20 KB patches): 500 ≈ well under 1s
- heavy monorepo (~50-100 KB patches): 500 ≈ 2-5s

Branches merged within a normal review-and-cleanup cycle sit well inside this. Anything older is `-D` territory.

## Scope

The cap is in the shared probe, so it applies to `wt list`'s status column, `wt remove`, `wt merge`, and `wt step prune` — anywhere `integration_reason` is called. Behavior change only when a branch was squash-merged *and* `git merge-tree` conflicts (the same files were modified again after the squash) *and* the default branch has advanced > 500 commits since the merge point. Narrow.

## Test plan

- Two new unit tests in `patch_id_cap_tests` using `git fast-import` (501-commit history built in milliseconds): under-cap finds the squash, over-cap bails despite the squash being in range. The pair pins the difference to the cap, not the topology.
- Existing squash-merge integration tests (`test_remove_squash_merged_*`, `test_prune_squash_merged_*`, `test_list_integrated_when_squash_merged_*`) keep small histories well under the cap and continue to pass — regression guard for the wiring.

Refs #1888.

@ortonomy — would you mind running this against the repo from #1888? The simplest check is `cargo install --git https://github.com/max-sixty/worktrunk --branch prune-speed wt` (or build locally), clear the disk cache to force a cold run (`rm -rf .git/wt/cache`), then time `wt step prune`. Most interested in whether the silent stretch is gone; if it isn't, `wt -vv step prune` writes `.git/wt/logs/trace.log` showing which `git` invocations are slow.

### DIFF
--- a/docs/content/remove.md
+++ b/docs/content/remove.md
@@ -46,7 +46,7 @@ Worktrunk checks six conditions (in order of cost):
 3. **No added changes** — Three-dot diff (`target...branch`) is empty. Shows `⊂`.
 4. **Trees match** — Branch tree SHA equals target tree SHA. Shows `⊂`.
 5. **Merge adds nothing** — Simulated merge produces the same tree as target. Handles squash-merged branches where target has advanced with changes to different files. Shows `⊂`.
-6. **Patch-id match** — Branch's entire diff matches a single squash-merge commit on target. Fallback for when the simulated merge conflicts because target later modified the same files the branch touched. Shows `⊂`.
+6. **Patch-id match** — Branch's entire diff matches a single squash-merge commit on target. Fallback for when the simulated merge conflicts because target later modified the same files the branch touched. Shows `⊂`. The default-branch walk is capped so a single check stays fast; a squash merge with hundreds of commits landed since the merge point falls outside the cap and needs `-D` to remove.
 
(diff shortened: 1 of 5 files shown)


max-sixty:
I'll merge even though I'm not sure it solves every problem

PR #2843, opened 2026-05-20
fix(shell): shell-correct (PowerShell + fish) escaping for the --execute payload

max-sixty:
## What

`wt switch -x <cmd> -- <args>` builds the `--execute` payload as a single shell-escaped string. When shell integration is active that string is written to the EXEC directive file and evaluated by the user's shell wrapper. wt escaped it with POSIX single-quote escaping (`shell_escape::unix::escape`) unconditionally — but two wrappers do not have POSIX single-quote semantics:

- **PowerShell** evaluates the EXEC file with `Invoke-Expression`; the POSIX `'\''` quoting idiom is invalid PowerShell, which wants embedded quotes doubled (`''`). This completes the PowerShell-escaping follow-up that #2815 explicitly deferred.
- **fish** evaluates it with `eval`; inside `'…'` fish treats `\` as an escape character (`\\`→`\`, `\'`→`'`), unlike POSIX where `\` is literal. A POSIX-escaped payload containing a backslash is mis-evaluated under fish — `a\\b` silently loses a backslash, and a trailing backslash makes fish abort the whole `eval` with a quote-balance parse error.

## Approach

Escaping now keys on the active directive shell. A `ShellEscapeMode` enum (`Literal` / `Posix` / `PowerShell` / `Fish`) replaces the old `shell_escape: bool` parameter on `expand_template`; `directive_shell_escape_mode()` reads the `WORKTRUNK_SHELL` env var (the PowerShell wrapper already exported it; the fish wrapper now does too). Only the `--execute` payload is shell-aware — hooks and aliases always run through `Cmd::shell` (`sh` / Git Bash) and stay POSIX; filesystem-path templates stay `Literal`.

## For reviewers

- `src/shell_exec.rs` — the new escaping section: `ShellEscapeMode`, `directive_shell_escape_mode()`, `powershell_escape()`, `fish_escape()`, `shell_escape_for()`.
- `src/output/global.rs` — `escape_legacy_cd` consolidated onto the same shell decision.
- `src/commands/worktree/switch.rs` — the `--execute` block escapes via the active-shell escaper.
- `templates/fish.fish` — exports `WORKTRUNK_SHELL=fish` so wt can identify the wrapper.
- The diff is large (~+1250/−210) but the substantive change is ~150 lines; the rest is the mechanical `bool`→enum migration across `expand_template` call sites plus `cargo fmt` reflow of the now-longer calls.

## Non-breaking

For every non-PowerShell, non-fish user, behavior is byte-identical: `WORKTRUNK_SHELL` absent ⇒ `Posix` ⇒ the previous `shell_escape::unix::escape`. The pre-existing POSIX snapshots are unchanged.

## Testing

`cargo run -- hook pre-merge --yes` clean (3790 tests). The fish escaper was round-tripped through the actual fish binary (4.7.1) — adversarial inputs (consecutive, leading, and trailing backslashes; embedded quotes; `$(…)`; backticks; newlines; empty) all deliver byte-exact, and the original bug reproduces without the fix. New integration tests cover both the new-protocol EXEC path and the legacy `cd` directive under `WORKTRUNK_SHELL=powershell` and `=fish`; unit tests cover `powershell_escape` and `fish_escape` directly.


max-sixty:
Merging this as the correctness fix. Recording a direction for follow-up, though: I think we should probably strip all of this escaping out rather than keep extending it shell-by-shell. Every wrapper we support is a different quoting dialect — POSIX `'\''`, PowerShell `''`, fish `\` escapes — and each one is a place to get it subtly wrong. The fish gap this PR fixes was itself a hole in the "POSIX everywhere" assumption that held until someone looked closely. The cleaner end state is to do no escaping at all: hand the `--execute` payload to the wrappers as data — the NUL-separated-argv "B2" option in the design doc — so there is no quoting round-trip in any shell. That deletes the whole class of bug instead of leaving us one more escaper to keep correct.

PR #2900, opened 2026-05-25
fix: lifecycle-safety follow-ups (#2870)

max-sixty:
Two follow-ups to #2870 surfaced by a retrospective deep-dive review. The
third item discussed in that review (the two divergent safe-delete semantics
in `execute_instant_removal_or_fallback`) is being explored as a separate
CAS-based rewrite via `git update-ref -d <ref> <sha>` and will come in a
follow-up PR — it touches `delete_branch_if_safe` core logic and warrants
its own review surface.

### 1. `refactor(prune): drop defensive check_lock now that phases are sequenced`

The `check_lock` `RwLock<()>` in `step_prune` was carried over from #2808's
Windows `.git/config` race fix. After #2870 restructured the flow so hook
approval is exact, the phase ordering already serializes integration-check
readers against the rename-failure fallback: the background `par_iter` is
the only reader, the channel-drain `for ... in rx` loop exits only once
that thread (and all its readers) has finished, and removals run strictly
after.

The remaining comment admitted the guard was defensive (\"preserves the
serialization if this flow is refactored back to overlap checks and
removals\") — one mechanism per guarantee per CLAUDE.md, so the lock is
removed. Comments on the par_iter spawn and the `SynchronousForNonCurrent`
variant now name the real invariants.

**Verdict:** Removed. Phase ordering is the actual mechanism; the lock was
belt-and-suspenders.

### 2. `fix(remove): warn when background branch deletion silently retains the branch`

Both branch-deletion paths in `execute_instant_removal_or_fallback`
(the fast-path post-prune delete and the `SynchronousForNonCurrent`
fallback delete) swallowed errors at `log::debug!` — invisible at default
verbosity. When the worktree was removed but the branch survived (a hook
advanced it, `git branch -D` errored, refs scan failed), the user got no
signal that the second half of the operation didn't happen.

Promoted to a `warning_message` with a `wt remove --force-delete <branch>`
recovery hint. To avoid duplicating the planner's existing \"Branch
unmerged\" hint, the warning suppresses the `Ok(NotDeleted)` case when
the planner already predicted retention; race-driven `NotDeleted` outcomes
still warn. `Err` always warns. `BackgroundRemoval` carries a new
`planner_expected_retention` bit so the helper can make that distinction
without re-deriving planner state.

Tradeoff: a previously silent path gains noise on the rare race. Silent
half-completion of a destructive operation is worse than an occasional
extra warning line.

### Tests

`cargo run -- hook pre-merge --yes` passes: 3834 nextest tests + doctests
+ pre-commit lints + clippy + docs.

### DIFF
--- a/src/commands/step/prune.rs
+++ b/src/commands/step/prune.rs
@@ -3,7 +3,7 @@
 use std::collections::HashSet;
 use std::fs;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Context;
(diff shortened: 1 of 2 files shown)


max-sixty:
Superseded by #2903, which already contains every commit from this PR plus the full atomic-CAS branch-deletion rework. Closing in favor of that; no work is lost.

PR #2902, opened 2026-05-25
refactor(log): replace env_logger with layered tracing-subscriber

max-sixty:
## Summary

Replace `log` + `env_logger` routing with `tracing` + `tracing-subscriber` layered subscribers. Routing of records to stderr / `trace.log` / `output.log` is now structural (per-layer `Filter`) rather than a per-record `Route` decision in a format closure.

- Three layers, three filters, three writers — each with a tiny `FormatEvent` impl that preserves the existing on-wire output byte-for-byte (styled stderr; plain-text trace.log; raw output.log).
- `log::*` callers untouched — `tracing_log::LogTracer` bridges them. `log::max_level` is pinned to the higher of `-v` baseline and `RUST_LOG`, so the static `log_enabled!` short-circuit still keeps unused format args from being evaluated.
- `[wt-trace]` grammar unchanged — `wt-perf`, the timeline benchmarks, `analyze_trace.rs` integration tests, and the diagnostic file all parse the same format. `trace::emit` switches to native `tracing::debug!` and `tracing::enabled!`.

## What got deleted

- `log_files::Route` enum and `route()` function
- The ~30-line `Builder::format(|buf, record| …)` closure in `init_logging`
- The `Repository::current()` priming dance in `init_logging` (subscriber order replaces it: open files → install subscriber → bridge log)
- `env_logger` dependency
- A 60-line block of doc comments explaining the old chicken-and-egg of file-open-before-vs-after env_logger registration

## What got added

- `src/logging.rs` (260 lines): three layer constructors, three `FormatEvent` impls, a `LogTracer` init that caps log::max_level to match the verbosity, and a `RUST_LOG` ceiling helper. All the routing logic in one place, with module-level docs.

Net: +259 / -232 lines of source (Cargo.lock unchanged net), but several conceptual artifacts removed.

## Behavior preserved (per `tests/integration_tests/diagnostic.rs`)

| Input | Effect |
|---|---|
| no flag, no env | nothing visible (Warn+ only) |
| `RUST_LOG=debug` | Debug to stderr; bounded preview only; no file sinks |
| `-v` | Info to stderr |
| `-vv` | log pipeline silent on stderr; structured + bounded preview to `trace.log`; raw bodies to `output.log`; one-line startup pointer on stderr |
| `SUBPROCESS_FULL_TARGET` | never to stderr, only output.log when active |

All 1761 integration tests + 676 unit tests + 27 doctests pass locally. Pre-merge clean.

## Notes for review

- The one rename — `OUTPUT_LOG_AVAILABLE` → `OUTPUT_LOG_ACTIVE` in `shell_exec` — is just accuracy: "available" was a vestige of a draft that considered modeling the state as `Option`; the elision marker, the only consumer, just wants to know if `output.log` is being written to.
- `LogTracer` defaults `log::max_level` to `LevelFilter::max()` (Trace). Without my explicit `with_max_level(…)`, every `log::debug!(…)` site evaluates its format args, exposing arithmetic that's safe today only because the macro short-circuits — e.g. `now_secs - cached.checked_at` in `list/ci_status` panics on a clock-skewed test fixture. The explicit cap keeps the previous semantics.
- `Option<Layer>` is itself a `Layer` (no-op when `None`), so verbosity gates compose naturally with subscriber `.with(...)`.

## Test plan

- [x] `cargo test --lib --bins` — 676 pass
- [x] `cargo test --test integration` — 1761 pass
- [x] `cargo run -- hook pre-merge --yes` — 3834 nextest pass, all lints + doctests clean
- [x] Manual `wt -vv list` — trace.log + output.log populated with expected content
- [x] Manual `RUST_LOG=debug wt list` — debug records on stderr with the `[a] $ git …` styling
- [ ] CI green (auto-checked)
- [ ] `codecov/patch` (auto-checked; if it fails, will investigate before merging)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1339,6 +1339,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
 
 [[package]]
+name = "matchers"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d1525a2a28c7f4fa0fc98bb91ae755d1e2d1505079e05539e35bc876b5d65ae9"
+dependencies = [
+ "regex-automata",
+]
+
+[[package]]
 name = "memchr"
 version = "2.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1452,6 +1461,15 @@ name = "normalize-path"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f5438dd2b2ff4c6df6e1ce22d825ed2fa93ee2922235cc45186991717f0a892d"
+
+[[package]]
+name = "nu-ansi-term"
+version = "0.50.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
+dependencies = [
+ "windows-sys 0.60.2",
+]
 
 [[package]]
 name = "num-conv"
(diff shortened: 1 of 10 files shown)


max-sixty:
Closing without GitHub-side merge — content landed on main as commit [`93812660c`](https://github.com/max-sixty/worktrunk/commit/93812660c) via `wt merge` (worktrunk's own land-my-work command), which squash-pushed directly to main and bypassed the PR close path. The squash SHA differs from this PR's head SHA, so GitHub didn't auto-close on push.

No code is missing — the tracing-subscriber migration is on `main`. Subsequent follow-ups #2906 (drop `OUTPUT_LOG_ACTIVE` static) and #2907 (structured `[wt-trace]` fields) have also landed.

PR #2941, opened 2026-05-30
feat(list): add repo_url to JSON output

bendrucker:
Adds a top-level `repo_url` to `wt list --format=json` and `wt list statusline --format=json`, derived locally from the primary remote. Consumers that want a link to the repository itself (not the PR) currently shell out to `git remote get-url` and convert SSH to HTTPS by hand; `repo_url` does that from the already-parsed remote.

Also adds `ci.repo_url`: the repository a PR/MR targets, derived from `ci.url`. In a fork checkout the primary remote is the contributor's fork while the PR lives upstream, so the two can differ — `repo_url` names the local repo, `ci.repo_url` names the PR target. I kept them as separate fields so neither has to guess which one the consumer meant.

## Notes

- `GitRemoteUrl::web_url()` owns the per-forge shape, including the Azure `dev.azure.com` / `*.visualstudio.com` split; it calls `azure::fork_remote_url` rather than duplicating the format.
- `repo_url_from_ref_url()` shares a tokenizer with the existing `parse_ref_url`, so PR/MR URL shapes live in one place.
- Both derive from local state — no network access, consistent with the local-first policy.


max-sixty:
looks good @bendrucker , thanks!

@worktrunk-bot anything we need to look over in particular? is Azure treated differently from others? is there a way to totally abstract out of the specific forge here?

max-sixty:
thank you @bendrucker !

merging as experimental, haven't done quite enough checking but trust you!

bendrucker:
If there are going to be edge cases I tend to assume they're most likely with the Azure repos. If any bugs come up, happy to work on them.

bendrucker:
Was going to follow up with accepting URL on `switch` only to find #2898 🙌🏻

max-sixty:
> Was going to follow up with accepting URL on `switch` only to find #2898 🙌🏻 

yes nice!

(though would it need a change for the Azure format?)

PR #2959, opened 2026-06-01
fix: prevent console storm on windows

nathanbabcock:
**Tl;dr:** Background steps during hooks like `[post-start]` briefly spawn visible consoles on Windows. The fix is to use `CREATE_NO_WINDOW`. Apart from suppressing the console windows, it functions identically.

---

From [**Process Creation Flags** (Microsoft Learn)](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags):

`DETACHED_PROCESS` — `0x00000008`

> For console processes, the new process does not inherit its parent's console (the default). The new process can call the [`AllocConsole`](https://learn.microsoft.com/en-us/windows/console/allocconsole) function at a later time to create a console. For more information, see [Creation of a Console](https://learn.microsoft.com/en-us/windows/console/creation-of-a-console). This value cannot be used with `CREATE_NEW_CONSOLE`.

`CREATE_NO_WINDOW` — `0x08000000`

> The process is a console application that is being run without a console window. Therefore, the console handle for the application is not set. This flag is ignored if the application is not a console application, or if it is used with either `CREATE_NEW_CONSOLE` or `DETACHED_PROCESS`.

---

Verified locally on a `wt.toml` that had 4 pre-start hooks (some semi long-running like a `pnpm install`)

<table><thead><tr><th>Before</th><th>After</th></tr></thead>
<tbody><tr><td>

[worktrunk-console-storm-before.webm](https://github.com/user-attachments/assets/4e88b798-e2d0-43a1-a750-da1eb41ece76)

</td><td>

[worktrunk-console-storm-after.webm](https://github.com/user-attachments/assets/d572cc6d-7d39-40ff-a851-afb5ff733972)

</td></tr></tbody></table>

### DIFF
--- a/src/commands/process.rs
+++ b/src/commands/process.rs
@@ -283,9 +283,9 @@ fn spawn_detached_windows(
     use worktrunk::shell_exec::ShellConfig;
 
     // CREATE_NEW_PROCESS_GROUP: Creates new process group (0x00000200)
-    // DETACHED_PROCESS: Creates process without console (0x00000008)
+    // CREATE_NO_WINDOW: Creates process without console window (0x08000000)
     const CREATE_NEW_PROCESS_GROUP: u32 = 0x00000200;
-    const DETACHED_PROCESS: u32 = 0x00000008;
+    const CREATE_NO_WINDOW: u32 = 0x08000000;
 
     let shell = ShellConfig::get()?;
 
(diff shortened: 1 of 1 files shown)


max-sixty:
thanks @nathanbabcock 

to be totally transparent, I'm quite green with Windows...

are you confident this is a good change? happy to merge if you are!

nathanbabcock:
@max-sixty yeah, I can vouch for it. It's a fix I've used before to reliably prevent consoles from popping up on Windows.

To clarify a bit more this specific case, it's actually the *grandchild* processes that are popping up.

The initial child process spawned by `wt` is Git Bash on Windows, and either flag works — both `DETACHED_PROCESS` and `CREATE_NO_WINDOW` will correctly prevent a console window from Git Bash itself.

The difference is when Git Bash spawns another binary that wants to produce console output (like almost any dev tool). In the `DETACHED_PROCESS` case, Git Bash has no console to pass along to its children, and the default behavior will be for those children to spawn their own consoles. With `CREATE_NO_WINDOW` on the other hand, Git Bash has an invisible console, and its children are able to inherit that one rather than creating their own.

max-sixty:
thank you v much!

PR #2978, opened 2026-06-04
fix(shell): tag zsh init script with #compdef so it works from $fpath

zetlen:
## Problem

The zsh integration script only works via `eval "$(wt config shell init zsh)"` in `.zshrc`. Completion managers that cache generated completion files (zinit, [zsh-completion-generators](https://github.com/zetlen/zsh-completion-generators), oh-my-zsh custom completions) instead save the output as `_wt` in a directory on `$fpath` — and compinit only registers files whose **first line** is a `#compdef` tag, so completions never load from there.

There's also a second-order issue: even with the tag added, the first TAB is a dud. compinit autoloads the file body as the completion function `_wt`; the body registers `compdef _wt_lazy_complete wt` for *future* completions but produces no candidates for the in-flight one.

## Fix

Two changes to `templates/zsh.zsh`, both inert in the existing `eval` flow:

- **`#compdef {{ cmd }}` as line 1** — an ordinary comment when eval'd; a registration tag when scanned by compinit. Same dual-use convention as clap_complete's own `COMPLETE=zsh wt` registration script and kubectl/docker completion files.
- **funcstack-guarded dispatch at the end** — when the body runs as the autoloaded completion function, complete the current word immediately:

  ```zsh
  if [[ "${funcstack[1]}" == "_wt" ]]; then
      _wt_lazy_complete "$@"
  fi
  ```

  At eval time `funcstack` is empty at the top level, so this never fires (verified — sourcing/eval'ing the script produces no spurious output).

## Verification

- New black-box test `test_zsh_completion_fpath_autoload` (gated on `shell-integration-tests`, alongside `test_zsh_completion_subcommands`): saves `wt config shell init zsh` output as `_wt` in a temp fpath dir, runs real compinit, asserts the completer is registered and the **first** invocation yields subcommands. Before the fix it snapshots `completer: NONE`; after, `completer: _wt` plus the subcommand list.
- `wt config shell init --help` documents the fpath usage.
- Snapshots refreshed (`init_zsh`, unit `init_zsh`); docs-sync test passes; fmt/clippy clean; full integration failure set is identical to `main` on my machine (remaining failures are local-env only: missing fish/nu, host shell-detection leaks).

### DIFF
--- a/src/cli/config.rs
+++ b/src/cli/config.rs
@@ -31,6 +31,8 @@ Zsh (~/.zshrc):
 eval "$(wt config shell init zsh)"
 ```
 
+The zsh script starts with a `#compdef` tag, so it also works as an fpath completion file: save it as `_wt` in a directory on `$fpath` before compinit runs, and completions load on first use. Completion managers that cache generated completion files can consume it this way.
+
 Nushell [experimental] — save to vendor autoload directory:
 ```console
 $ wt config shell init nu | save -f ($nu.default-config-dir | path join vendor/autoload/wt.nu)
--- a/src/shell/snapshots/worktrunk__shell__tests__init_zsh.snap
+++ b/src/shell/snapshots/worktrunk__shell__tests__init_zsh.snap
@@ -2,7 +2,13 @@
 source: src/shell/mod.rs
 expression: output
 ---
+#compdef wt
 # worktrunk shell integration for zsh
+#
+# Loads two ways:
+#   eval "$(wt config shell init zsh)" in .zshrc, or saved as a
+#   completion file (e.g. _wt in a $fpath directory) — the #compdef tag
+#   above lets compinit autoload it on the first completion.
 #
 # Completions require zsh's completion system (compinit). If completions don't work:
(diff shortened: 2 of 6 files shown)


max-sixty:
thanks a lot @zetlen !! this looks as high quality as I know zsh...

@worktrunk-bot given I don't know zsh, is there anything worth flagging at _all_? this is obv a really sensitive piece of code, and folks have very heterogeneous setups, so we should think through all the cases we can...

max-sixty:
OK re the clarification; and to confirm, our standard install still installs the shell integration?

max-sixty:
Ok great.

@zetlen if you could give this a once-over, given how important the file is for the project, that would be great, then we can merge. Thank you!

zetlen:
It's a good catch, actually--I hadn't noticed that there's also a chdir hook. I actually think that makes my change unwise. I'm pulling `wt` out of my zsh-completions-generator since using it on the fpath alone actually breaks the complete shell integration. Thanks!

PR #2983, opened 2026-06-05
Expose commit bodies to squash templates

### LINKED ISSUE #2982: Preserve commit message context in squash prompt templates
In larger codebases, depending on the workflow, a branch can collect small commits from teammates or coding agents. The subject usually captures the immediate change. The body carries the context future readers care about: why the change exists, what problem it solves and which constraints shaped the implementation.

For example:

```
|
* commit A
|
|     Strip location metadata from uploaded photos
|
|     Remove GPS metadata before storing originals so shared albums do not
|     expose where a photo was taken.
|
|     Refs: #1842
|
* commit B
|
|     Preserve image orientation after metadata removal
|
|     Apply the orientation before stripping EXIF data so portrait photos do
|     not appear rotated.
|
* commit C
|
|     Add regression test for portrait uploads
|
|     Verify portrait photos still render upright after metadata is removed.
```

Currently, custom squash templates can only include the subjects:

```
- Strip location metadata from uploaded photos
- Preserve image orientation after metadata removal
- Add regression test for portrait uploads
```

That leaves the LLM to reconstruct intent from the diff. It may produce a squash message that sounds plausible but drifts from the intended emphasis or stays generic:

```
feat(uploads): update photo metadata handling

Strip location metadata from uploaded photos, preserve image orientation,
and add regression coverage for portrait uploads.
```

When the commit bodies carry that intent the squash prompt can preserve it more directly:

```
feat(uploads): remove photo location metadata

Remove GPS metadata from uploaded photos before storing originals so
shared albums do not expose where a photo was taken. Preserve image
orientation so portrait uploads still render upright after metadata is
removed.

Refs: #1842
```

## Proposal

One possible approach would be to expose commit bodies to squash templates without changing the existing `commits` variable, so existing templates stay compatible.

For example, the squash template could have access to each commit's subject and body. That would let users pass the context they care about to their LLM.

To keep the scope small, this likely only needs the local message content already present in the commits, not authors, hashes, dates or parsed trailers.

florianilch:
Closes #2982

This addresses the issue with an additive template variable, leaving `commits` unchanged.

- Adds `commit_details` to squash prompt templates as a list of `{ subject, body }`.
- Derives the existing subject-only `commits` value from `commit_details`.
- Uses the existing repository command to collect bodies with `git log` `%b` and NUL delimiters to preserve multiline bodies.
- Documents the variable and adds unit/integration coverage.

## Example

After this change, squash templates can include commit bodies:

```jinja
{% for c in commit_details -%}
- {{ c.subject }}
{% if c.body -%}
{{ c.body | trim | indent(2, true) }}
{% endif -%}
{% endfor %}
```

## Verification

```bash
cargo test -p worktrunk commit_message_details
cargo test -p worktrunk llm::tests
cargo test --test integration test_step_squash_show_prompt_custom_template_renders_commit_details
```

### DIFF
--- a/dev/config.example.toml
+++ b/dev/config.example.toml
@@ -250,7 +250,7 @@
 # Available variables (in addition to commit template variables):
 #
 # - `{{ commits }}` — list of commit subjects being squashed
-# - `{{ commit_details }}` — list of commits being squashed as `{ subject, body }` objects
+# - `{{ commit_details }}` — [experimental] list of commits being squashed as `{ subject, body }` objects
 # - `{{ target_branch }}` — merge target branch
 #
 # Default template:
--- a/dev/config.example.toml
+++ b/dev/config.example.toml
@@ -249,7 +249,8 @@
 #
 # Available variables (in addition to commit template variables):
 #
-# - `{{ commits }}` — list of commits being squashed
+# - `{{ commits }}` — list of commit subjects being squashed
+# - `{{ commit_details }}` — list of commits being squashed as `{ subject, body }` objects
 # - `{{ target_branch }}` — merge target branch
 #
 # Default template:
--- a/src/git/repository/diff.rs
+++ b/src/git/repository/diff.rs
@@ -7,6 +7,37 @@ use dashmap::mapref::entry::Entry;
 
 use super::{DiffStats, LineDiff, Repository};
 
+/// Subject and body for one commit in a range.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize)]
+pub struct CommitMessageDetail {
+    /// Commit subject from git's `%s` pretty format.
+    pub subject: String,
+    /// Commit body from git's `%b` pretty format, including trailer-like lines.
+    pub body: String,
+}
+
+fn parse_commit_message_details_output(output: &str) -> anyhow::Result<Vec<CommitMessageDetail>> {
+    if output.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    let parts = output.split('\0').collect::<Vec<_>>();
+    if parts.len() % 2 != 0 {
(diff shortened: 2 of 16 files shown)


src/git/repository/diff.rs
```diff
@@ -112,16 +143,21 @@ impl Repository {
...
+        // Git pretty-format placeholders:
+        // - `%s`: subject
+        // - `%x00`: literal NUL delimiter
+        // - `%b`: body as Git reports it; trailer-like lines remain part of this text
         let output = self.run_command(&[
             "log",
+            "-z",
             "--no-show-signature",
-            "--format=%s",
+            "--pretty=format:%s%x00%b",
```
max-sixty, line 156: @worktrunk-bot are there any issues about color here? (Guessing not but checking)

dev/config.example.toml
```diff
@@ -249,7 +249,8 @@
 #
 # Available variables (in addition to commit template variables):
 #
-# - `{{ commits }}` — list of commits being squashed
+# - `{{ commits }}` — list of commit subjects being squashed
+# - `{{ commit_details }}` — list of commits being squashed as `{ subject, body }` objects
```
max-sixty: I guess we should have named the original `commit_subjects`, so this could be named `commits`

Possibly we deprecate the old one and then move this over, but in another PR

max-sixty (approved):
Can we mark as experimental @worktrunk-bot ?

max-sixty:
@worktrunk-bot please start a deprecation cycle:
- move `commits` to deprecated
- usethe property of `commit_details` in its place, updating docs and the update command
- open an issue to, in three months, turn `commits` into an error
- then at some point in the future we'll switch to using `commits` to the dict

PR #3014, opened 2026-06-08
git: avoid racy temp-index stats for untracked files

Xeonacid:
`working_tree_diff_stats_with_untracked()` used a copied index to make untracked files visible to git diff: first `git add --intent-to-add .`, then `git diff --shortstat HEAD` against that temporary index. That means the copied index also has to rediscover tracked working-tree modifications.

On a slow riscv64 machine, the test passes by itself and the full lib test suite passes with `--test-threads=1`, but the default concurrent lib test run reliably exposes a failure. Diagnostic output from the failing test showed tracked.txt already contained the new contents, while git status and git diff through the temporary index reported only untracked.txt. The tracked rewrite was same-size, so the copied index could treat it as racily clean under this slow riscv64 machine.

Do not use the temporary index to rediscover tracked changes. Keep the existing `git diff --shortstat HEAD` path for tracked, staged, and unstaged changes, then list untracked files with `git ls-files --others --exclude-standard -z`, stage only those paths into the temporary index, and sum their cached numstat output. The real index remains untouched, while the racy-index window is avoided.

### DIFF
--- a/src/git/repository/working_tree.rs
+++ b/src/git/repository/working_tree.rs
@@ -10,6 +10,7 @@ use dunce::canonicalize;
 
 use super::{GitError, LineDiff, Repository};
 use crate::git::CommandError;
+use crate::git::parse_numstat_line;
 
 /// Parse `git submodule status` output and detect whether any submodule is initialized.
 ///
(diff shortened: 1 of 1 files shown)


max-sixty:
> On a slow riscv64 machine, the test passes by itself and the full lib test suite passes with --test-threads=1, but the default concurrent lib test run reliably exposes a failure. Diagnostic output from the failing test showed tracked.txt already contained the new contents, while git status and git diff through the temporary index reported only untracked.txt. The tracked rewrite was same-size, so the copied index could treat it as racily clean under this slow riscv64 machine.

do we know where this race comes from?

@worktrunk-bot please can you explain it to me?

max-sixty:
thanks!

if we copied the index and maintained its timestamp, would that work?

Xeonacid:
Thanks for the great review and explanation, max-sixty and bot!

> Why only slow riscv64-under-qemu

Nit: The bug occurs not only under qemu, but also on real riscv64 hardware which are mostly slow.

max-sixty:
@Xeonacid thank you _very_ much for the thoughtful and technically deep change!

PR #3087, opened 2026-06-15
fix(picker): stop first-keystroke freeze

bendrucker:
Fixes a multi-second freeze in the `wt switch` picker: with many accumulated worktrees, typing the first character locks the UI for seconds, then it recovers. The freeze scales with worktree count.

The picker (skim) runs its per-keystroke fuzzy matcher and result sort on rayon's **global** thread pool. Worktrunk's collection floods that same global pool with blocking git subprocess tasks (status, diff, rev-list, merge-base, plus the preview orchestrator's per-mode `git diff` / `git log`), one batch per worktree. The global pool has only `2× CPU` workers, and each git call blocks its worker for the subprocess lifetime. When the user types, skim's matcher queues behind that flood and can't run until workers drain. `wt list` never froze because nothing else contends for the pool there.

The fix moves the git-heavy collection and preview work onto a dedicated `COLLECT_POOL`, leaving the global pool free for skim. This is the same isolation pattern already used by `copy::COPY_POOL` and `remove_dir::REMOVE_POOL`. The new pool is sized like the global pool (`2× CPU`, honoring `RAYON_NUM_THREADS`), so collection throughput is unchanged. Its only job is to keep the git work off the pool skim's matcher uses.

## Decisions

- Collection's row pipeline and the preview orchestrator stay on the same dedicated pool, preserving the orchestrator's intentional "one shared pool, let workers prefer dominant pressure" design. They just move off the pool skim needs.
- The bounded pre- and post-skeleton `rayon::scope` calls stay on the global pool. They are O(1) in worktree count (~7 spawns), so they are not the scaling flood.
- The single-item statusline path (`populate_item`) routes through `COLLECT_POOL` too, purely for consistency. A single item never floods the pool, so this path was never the problem.
- The nested log-refresh `rayon::spawn_fifo` needs no change. The free `rayon::spawn_fifo` resolves its target via the current worker's registry, so when called from inside a `COLLECT_POOL` worker it inherits `COLLECT_POOL` rather than falling back to the global pool. Confirmed against the rayon-core source.

## Testing

Ran locally, before and after below, depicting the freeze/fix

### Before

https://github.com/user-attachments/assets/55c12cec-4466-487a-b480-fd2ff03ad111

### After

https://github.com/user-attachments/assets/71a5499a-5521-4727-a6b4-ab6a12f317d5

### DIFF
--- a/src/commands/list/collect/mod.rs
+++ b/src/commands/list/collect/mod.rs
@@ -337,12 +337,11 @@ pub(crate) static COLLECT_POOL: LazyLock<rayon::ThreadPool> = LazyLock::new(|| {
         .expect("failed to build collect thread pool")
 });
 
-/// The `num_threads` argument for [`COLLECT_POOL`]. Mirrors `main.rs`'s
-/// global-pool sizing: when `RAYON_NUM_THREADS` is set, return 0 so Rayon
-/// reads and validates the env var itself; otherwise return
-/// [`crate::rayon_thread_count`] (`2× CPU`), since Rayon's own default is
-/// `1× CPU`. Takes the env presence as a parameter so both branches are
-/// unit-testable without mutating process-global environment.
+/// The `num_threads` argument for [`COLLECT_POOL`]. When `RAYON_NUM_THREADS`
+/// is set, return 0 so Rayon reads and validates the env var itself.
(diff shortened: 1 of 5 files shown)


max-sixty:
nice! thanks a lot @bendrucker !

PR #3100, opened 2026-06-17
Release v0.59.0

### LINKED ISSUE #2926: wt switch file picker becomes sluggish during background sync on large repositories
## Overview

When using `wt switch` in larger repos with extensive git history, the interactive branch picker becomes noticeably slow and laggy while background sync tasks are running. The issue is most apparent when trying to type into the search/filter input after the picker opens. Keystrokes are delayed and the UI struggles to keep up until the background work finishes.

This makes branch switching frustrating in larger repos where responsive filtering is especially important. However I've noticed in smaller repos it's not a problem.

Repo size
```
❯ du -sh .git
2.3G    .git
❯ git rev-list --all --count
76369
```

## Steps to Reproduce

1. Open a large git repository with substantial commit history and many branches
2. Run:

`wt switch`

3. Wait for the branch/file picker UI to appear
4. Immediately begin typing to filter/search branches
5. Observe input lag and sluggish UI responsiveness while background sync operations are still running

### Expected Behavior

The picker UI and search/filter input should remain responsive even while background sync tasks are executing.

### Actual Behavior

The search/filter input becomes slow and laggy during background sync activity. Typing responsiveness improves once the sync work completes.

## Additional Notes

* This seems correlated with repository size and git history depth
* The slowdown appears specifically tied to concurrent background sync activity
* Feels like the sync work may be competing with the interactive UI thread/event loop

max-sixty:
Release v0.59.0 (minor). The bump is minor because two backward-incompatible behavior changes ship: in the `wt switch` picker, `1`–`5` no longer switch preview tabs (digits now filter; tabs move to `Alt-1`–`Alt-5` / `Tab`); and `wt step eval --dry-run` is removed in favor of `-v`. `cargo semver-checks` reports no library API breakage.

### Highlights

- **Improved**: picker frees digit keys for filtering (preview tabs to `Alt`); squash templates use `commit_details` instead of the deprecated `commits`; `wt step eval -v` lists template variables.
- **Fixed**: picker no longer freezes on the first keystroke with many worktrees (#3087, thanks @bendrucker; fixes #2926, thanks @mahume for reporting); tab-completion covers all hook types; Ctrl-C on a concurrent alias reports the right exit code.

See `CHANGELOG.md` for the full set.

### Bundled test fix

This PR also carries one non-release commit: `test(ci_status): cover '?' worker-thread label in snapshot filter`. The `test_list_full_with_invalid_configured_platform` snapshot filter only normalized `[a-zA-Z]` worker-thread labels, but `label_for_thread_index` emits `?` for thread IDs above 52. On high-core machines the rayon pools span enough threads to hit that, so the gate failed locally (CI's low core count never triggers it). The filter now covers the full label alphabet. Test-only; not in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Changelog
 
+## 0.59.0
+
+### Improved
+
+- **Picker frees digit keys for filtering; preview tabs move to `Alt`**: In the `wt switch` picker, plain digits now go to the filter, so a branch name with a number in it can be typed directly. Preview tabs jump with `Alt-1`–`Alt-5` or cycle with `Tab`/`Shift-Tab`. (Breaking: `1`–`5` no longer switch preview tabs) ([#3079](https://github.com/max-sixty/worktrunk/pull/3079))
+
+- **Squash templates use `commit_details` instead of `commits`**: The squash commit-message template's `{{ commits }}` variable (commit subjects) is deprecated in favor of `{{ commit_details }}`, which renders as the bare subject when printed directly and also exposes `.subject` and `.body`. `wt config update` migrates `commits` to `commit_details` as a plain rename. ([#2985](https://github.com/max-sixty/worktrunk/pull/2985))
+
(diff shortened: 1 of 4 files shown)


mahume:
Just did a manual test for `wt switch`. Looks good!

PR #3105, opened 2026-06-17
feat(completions): ship dynamic registration for package installs

bendrucker:
A plain package install of `wt` cannot complete branch or worktree names.

```
brew install worktrunk
# new shell
wt switch <TAB>     # subcommands and flags only, no branch names
```

I found that `wt config shell completions <shell>` emits a static `clap_complete::generate()` registration, and the homebrew-core formula bakes that into `site-functions/_wt` at install time via `generate_completions_from_executable`. The 0.58.0 binary shows:

```
$ wt config shell completions zsh | grep -c _arguments   # static _arguments calls
88
$ wt config shell completions zsh | grep -c COMPLETE     # dynamic runtime markers
0
```

That static `_wt` is what every Homebrew user has autoloaded by `compinit`. It completes subcommands and flags and has no mechanism to list branches.

## Proposed Fix

`wt config shell completions` now emits a **dynamic** registration, the same script `COMPLETE=<shell> wt` already prints, which calls the binary at TAB time. The binary has always supported dynamic completion at runtime. The package path was simply wired to the wrong clap API: `clap_complete::generate()` is the legacy static path, and `clap_complete::env` is the maintained dynamic one. After this change a plain `brew install` completes branch and worktree names with zero config, matching what `gh`, `rustup`, `kubectl`, and `docker` ship via their packages.

## Changes

- **Bash, Zsh, Fish, and PowerShell now emit dynamic registration** via `clap_complete::env`. It reuses clap's maintained body rather than hand-rolling a completer. Nushell already emitted its full dynamic template, so it is unchanged.
- **`make_zsh_autoload_safe`** rewrites clap's trailing `compdef` into a dual-mode guard. Homebrew installs the zsh script as `site-functions/_wt` and autoloads it from `fpath`, where the file body itself runs as the completion function. The guard completes directly when autoloaded and falls back to `compdef` when sourced. The display zstyles precede the guard so single-column listing applies on the first TAB.
- **The other shells need no transform.** Homebrew sources bash files, autoloads fish completion files by command name, and PowerShell registrations are sourced into the profile.
- **Dropped a dead `sed` hack** from `templates/zsh.zsh`. clap emits `_describe -V` natively, so the substitution matched nothing.
- **Reworded the `Completions` help text**, which falsely claimed "static / only commands and flags."

Homebrew installs completions for bash, zsh, and fish. PowerShell is switched to dynamic for consistency, but no package manager installs its completions, so it applies only when a user sources it.

## Why Not `wt config shell install`?

`install` does bind dynamic completion, but a package install never runs it and should not have to. `install` exists for cd-on-switch: its job is the `wt()` wrapper that lets `wt switch` change the shell's directory, and it does so by appending an `eval "$(...)"` line to `~/.zshrc` or `~/.bashrc`. Completion is incidental to it.

It only makes sense for Homebrew to run the completions command, since it's a pure function, no side effects. Changing the user's shell config during install would be unexpected.

## Testing

Tested real shells with the completer present *only* via `fpath` autoload (no eval, no install):

- zsh: `wt switch <TAB>` lists branch names on the **first** TAB, single-column.
- bash (sourced, as Homebrew does): branches appear.
- Dynamic value path (`COMPLETE=zsh … wt switch ''`) still lists branches with recency markers.

Fish reuses the exact `COMPLETE=fish wt` mechanism that `wt config shell install` already ships, so it is proven in-tree. The strengthened `test_completions_for_all_shells` pins the dynamic registration structure for every shell.

## Related

I hit this and worked around it locally by binding the dynamic completer by hand after `compinit` ([bendrucker/dotfiles#493](https://github.com/bendrucker/dotfiles/pull/493)).

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -1,11 +1,10 @@
 use anyhow::Context;
-use clap::CommandFactory;
-use clap_complete::generate;
+use clap_complete::env::{
+    Bash as EnvBash, EnvCompleter, Fish as EnvFish, Powershell as EnvPowershell, Zsh as EnvZsh,
+};
 use std::io::{self, Write};
 use worktrunk::shell;
 use worktrunk::styling::println;
-
-use crate::cli::Cli;
 
 pub fn handle_init(shell: shell::Shell, cmd: String) -> Result<(), String> {
     let init = shell::ShellInit::with_prefix(shell, cmd);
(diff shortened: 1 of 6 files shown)


bendrucker:
Looking into the test fail

max-sixty:
thank you very much @bendrucker !

so I understand: does this only change behavior for those who install with brew and _don't_ install the shell integration?

(or @worktrunk-bot feel free to opine)

src/commands/init.rs
```diff
@@ -71,14 +75,47 @@ pub fn handle_completions(shell: shell::Shell) -> anyhow::Result<()> {
...
+/// styles before the completer call means `_describe` sees them on the *first* TAB, not
+/// only on the second.
+fn make_zsh_autoload_safe(script: &str, cmd_name: &str) -> String {
+    let func = format!("_clap_dynamic_completer_{}", cmd_name.replace('-', "_"));
+    let trailing = format!("compdef {func} {cmd_name}");
+    let replacement = format!(
+        r#"# Single-column display keeps descriptions visually associated with each branch.
+zstyle ':completion:*:{cmd_name}:*' list-max 1
+# Prevent grouping branches with identical descriptions (same timestamp) on one line.
+zstyle ':completion:*:*:{cmd_name}:*' list-grouped false
```
max-sixty, line 112: @worktrunk-bot confirm whether these are necessary

max-sixty:
ok, great! it seems like a nice improvement + some simplification, even if most folks have the shell integration installed?

bendrucker:
Yep, I of course have the shell integration installed but found the dynamic completions were clobbered by the presence of the static ones from Homebrew.

max-sixty:
> Yep, I of course have the shell integration installed but found the dynamic completions were clobbered by the presence of the static ones from Homebrew.

ah!! that seems bad indeed...

are you on zsh? I wonder if it's some ordering thing there. 

regardless: now fixed!

bendrucker:
Yes, zsh, and yes related to ordering.

PR #3117, opened 2026-06-18
fix(switch): show correct worktree path in bare repo hidden-dir warning

ammachado:
## Summary

The net change over `main` is that the warning's paths now render in **full `~/…` form** instead of just their last two components — so the warning matches the `Created … @ …` success line below it. The which-path-goes-where semantics are unchanged from `main`:

- **Warning** (`worktrees will be at …`) shows the **hidden-dir destination** (`~/…/project/.git.feature`) — where worktrees actually land *without* the fix, matching the `Created @ …` line.
- **Hint / prompt** (`To place worktrees at …`) shows the **fixed sibling path** (`~/…/project/feature`) the config override would produce.
- Both lines, and "Bare repo at", now route through `format_path_for_display` for consistent full-path rendering.
- `example_bad` (the hidden-dir destination) and `example_good` (the sibling path) are both still computed — `example_bad` feeds the warning, `example_good` the hint/prompt.
- Also adds `.idea/` to `.gitignore` for RustRover (unrelated to the fix).

> Note: an earlier commit briefly flipped "worktrees will be at" to the *good* sibling path, but it was reverted in review (`fix(switch): warning shows current bad path, prompt shows fixed good path`) because it contradicted the `Created @ …/.git.feature` line on the decline / `--yes` / non-interactive paths. Net-zero on that axis; only the full-path rendering survives.

## Before / After

**Before (`main`):**
```
▲ Bare repo at project/.git — worktrees will be at project/.git.feature
↳ To place worktrees at project/feature, add to test-config.toml:
✓ Created branch feature and worktree @ [TEMP]/project/.git.feature
```

**After (this PR):**
```
▲ Bare repo at [TEMP]/project/.git — worktrees will be at [TEMP]/project/.git.feature
↳ To place worktrees at [TEMP]/project/feature, add to [TEMP]/test-config.toml:
✓ Created branch feature and worktree @ [TEMP]/project/.git.feature
```

(`[TEMP]` is the test harness substituting `$TMPDIR`; for a real user this renders as `~/…`. The warning and the `Created @ …` line now show the same full path.)

## Test plan

- [x] `cargo insta test --accept --test integration --features shell-integration-tests` — regenerates all five snapshots, including the three `shell-integration-tests` PTY ones
- [x] `cargo test --test integration bare_repo` — all bare-repo tests pass

### DIFF
--- a/src/commands/worktree/resolve.rs
+++ b/src/commands/worktree/resolve.rs
@@ -228,23 +228,29 @@ pub fn offer_bare_repo_worktree_path_fix(
         return Ok(false);
     }
 
-    // Display names for messages
+    // Display names for messages. `display_path` is the parent directory shown
+    // in the "Set worktree-path for …" success message. `bare_repo_display` is
+    // the repo itself, used in the warning ("Bare repo at …").
     let display_path = repo_path
         .parent()
         .map(|p| format_path_for_display(p).to_string())
         .unwrap_or_else(|| format_path_for_display(repo_path).to_string());
-    let parent_name = repo_path
-        .parent()
-        .and_then(|p| p.file_name())
-        .and_then(|n| n.to_str())
-        .unwrap_or("project");
+    let bare_repo_display = format!("{display_path}/{repo_name}");
 
     // Use the actual branch being switched to in example paths. Calling
     // `sanitize_branch_name` keeps the displayed example aligned with what
     // the `sanitize` Jinja filter applies inside `BARE_REPO_WORKTREE_PATH`.
     let sanitized = worktrunk::config::sanitize_branch_name(branch);
-    let example_bad = format!("{parent_name}/{repo_name}.{sanitized}");
-    let example_good = format!("{parent_name}/{sanitized}");
+
+    // The correct path is repo_path/../<sanitized-branch>: one level up from
+    // the bare-repo directory so worktrees sit as siblings (e.g. project/main
+    // rather than project/.git.main). This is both shown in the warning and
+    // offered as the destination in the prompt.
+    let good_path = repo_path
+        .parent()
+        .map(|p| p.join(&sanitized))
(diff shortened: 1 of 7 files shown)


ammachado:
All three points addressed:

1. PTY snapshots regenerated with `--features shell-integration-tests`.
2. Warning now shows the actual unconfigured destination (`…/.git.feature`); the hint/prompt shows the fixed path (`…/feature`). No more contradiction.
3. `.idea/` noted — kept as a convenience since it was already in the branch.

max-sixty:
@worktrunk-bot  please can you explain the change? use the snapshotsas examples. the only change I can see when looking quickly is we've maybe amde the paths absolute?

src/commands/worktree/resolve.rs
```diff
@@ -228,23 +228,37 @@ pub fn offer_bare_repo_worktree_path_fix(
...
+        .map(|p| p.join(&sanitized))
+        .unwrap_or_else(|| repo_path.join(&sanitized));
+    let example_good = format_path_for_display(&good_path).to_string();
+
+    // Where worktrees actually land *without* the fix, under the default
+    // `{{ repo }}.{{ branch }}` template — shown in the warning so it matches
+    // the "Created … @ …" line below instead of contradicting it.
+    let bad_path = repo_path
+        .parent()
+        .map(|p| p.join(format!("{repo_name}.{sanitized}")))
```
max-sixty, line 259: but isn't this reliant on the worktrees path config??

max-sixty:
> The bare repo hidden-dir warning now shows the resolved sibling path (e.g. ~/…/project/main) for "worktrees will be at", instead of the hardcoded broken-template path (project/.git.main)

@worktrunk-bot is this not accurate?

max-sixty:
ok, @ammachado feel free to start afresh if there is still an issue

PR #3142, opened 2026-06-20
feat(switch): make the picker filter text match the gutter sigil

max-sixty:
## What

In the interactive `wt switch` picker, each row carries a 2-cell gutter sigil — `@` current worktree, `^` primary worktree, `+` other/linked worktree, `/` local branch, `|` remote branch, `#` a `--prs` PR/MR row. Typing into the picker filtered only on each row's branch name + path (its skim `text()` / `search_text`), which deliberately excluded the gutter. So typing `+` matched nothing, even though every linked worktree shows a `+`.

This folds each row's gutter glyph into the fuzzy-search text, so typing a sigil filters by row kind: `+` narrows to linked worktrees, `@` to the current worktree, and `#` to `--prs` rows.

## How

A new canonical `ItemKind::gutter_glyph()` (and `BranchScope::gutter_glyph()`, replacing the old `gutter_sigil()`) is the single source of truth for both the rendered Gutter column and the picker's search text — so the displayed glyph and the searchable glyph can't drift. `WorktreeSkimItem` (worktree/branch rows) and `PrSkimItem` (`--prs` rows) each append their glyph to `search_text`. The render refactor is byte-identical (existing snapshots unchanged).

**Rank stability.** The glyph is a skeleton-time fact: `is_current`/`is_main` are set at item construction and `BranchScope` is structural (`is_previous`, set post-skeleton, never changes the glyph — previous worktrees share `+`). Folding it in therefore keeps skim's fuzzy ranks stable across the picker's progressive column updates, which was the constraint the existing `search_text` comment guarded.

## The `^` / `|` collision

The picker uses skim's default extended-search mode, where some characters are query operators. Confirmed from skim 0.20.5 source (`engine/factory.rs`, `engine/andor.rs`):

- `+`, `@`, `#`, `/` are literal — they match in the haystack.
- `^` is the prefix anchor: a bare `^` is an empty-pattern prefix that **matches every row**.
- `|` is the OR separator: a bare `|` trims to an empty query that **matches nothing**.

So `^` and `|` can't filter to the primary-worktree / remote-branch rows, and `/` over-matches because every path contains it. Rather than special-case the "good" glyphs, every row folds in its actual glyph uniformly, and the picker help documents which ones filter cleanly (`+`/`@`/`#`). An engine-level test (`gutter_glyphs_filter_under_skims_default_engine`) pins these semantics against skim's exact default factory stack, so a skim upgrade that changed them would fail loudly.

## Verified end-to-end

Driven against a real multi-worktree repo through the actual picker:

- `+` → exactly the linked worktrees
- `@` → exactly the current worktree
- `^` → all rows (collision, as documented)
- `|` → no rows (collision, as documented)

This live run also caught a doc bug in an earlier draft (the help had claimed `/` filters to local branches — it actually over-matches), now corrected.

## Testing

- `ItemKind::gutter_glyph` mapping per row kind.
- `WorktreeSkimItem` / `PrSkimItem` `search_text` includes the glyph.
- Engine-level test pinning the literal-vs-operator semantics.
- `cargo run -- hook pre-merge --yes` green (4038 tests; fmt + clippy clean; docs in sync).

## Note on the base

This stacks on #3128 (`switch-prs-picker`) — it edits `--prs` code (`PrSkimItem`) that only exists there. The diff below currently includes #3128's commits because `origin/switch-prs-picker` is behind the local tip this branch was cut from; it narrows to the single commit `7bbd804c5` once `switch-prs-picker` is pushed. A sibling branch is separately reworking the picker's CI-status path and touches the same two files (`items.rs`, `progressive_handler.rs`) but not `search_text` or the gutter — expect a trivial rebase when both land.


max-sixty:
Closing as superseded. The gutter-filter change on this branch landed on `main` directly via #3143 (the port) and #3154 (the `WorktreePresence` refactor), and the one piece unique to this branch — the `#` PR-row arm — is being folded into #3128 (`switch-prs-picker`) directly. Nothing here is lost.

PR #3172, opened 2026-06-23
feat(completion): mirror wrapped built-in for alias arg completion

yzx9:
An alias that forwards `{{ args }}` to a wt command — e.g. `co = "wt switch {{ args }}"` or `cm = "wt step commit {{ args }}"` — now inherits that command's argument and flag completion instead of the generic `--dry-run`/`--yes`/`--var` stub. Alias name completion already worked; this adds argument and flag completion.

Detection: the single command that forwards `{{ args }}` (via `template_references_var` — minijinja, not a substring, and scoped per command rather than the cross-command union), a leading `wt`, then a tree-walk `find_subcommand` that lands on a leaf subcommand. The leaf `Command` is cloned and renamed to the alias (clap `Command`/`Arg`/ `Extensions` and `ArgValueCompleter` are all `Clone`; the completer is `Arc`-backed, so the live git query survives the clone). Bare dispatchers (`wt step {{ args }}`), multiple `{{ args }}` forwarders, and aliases that do not forward args fall back to the generic stub.

Mirroring runs at the top level only; `wt step <alias>` keeps the generic stub.

### DIFF
--- a/docs/content/extending.md
+++ b/docs/content/extending.md
@@ -88,6 +88,8 @@ For indexing (`{{ args[0] }}`), looping, and counting, see [Passing values](@/ho
 
 Tokens after `--` forward unconditionally, bypassing any binding. Writing `wt deploy -- --branch=foo` forwards the literal `--branch=foo` to `{{ args }}` even though the template references `{{ branch }}`.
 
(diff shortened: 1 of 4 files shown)


max-sixty:
@worktrunk-bot ?

max-sixty:
very nice @yzx9 , thanks!

yzx9:
@max-sixty Thanks! Glad I could help. Great work on this!

PR #3196, opened 2026-06-24
feat(step): add require-include option + CLI flags to copy-ignored

yzx9:
`wt step copy-ignored` copies all gitignored files by default. Add a `require-include` setting that gates the copy on a `.worktreeinclude` file existing in the source worktree — matching Claude Code desktop, where the file is required. Without `.worktreeinclude`, the command is a no-op that reports why (text hint + JSON `reason`); with it, only matching files copy.

Config (additive, defaults off; `Option<bool>` merged via `.or()` so an explicit value overrides lower-priority config in either direction):

```toml
[step.copy-ignored]
require-include = true
```

CLI flags override config per run (precedence `CLI > --config-set > project > user > default`, mirroring `wt merge` / `wt remove`:

```
--require-include      copy nothing unless .worktreeinclude exists
--no-require-include   copy all gitignored files this run
```

The gate short-circuits before `git ls-files` discovery. The recovery hint names `--no-require-include` (correct whether the flag or config triggered it, since the flag overrides both). `wt step promote` is unaffected — it exchanges all gitignored artifacts by design.

### DIFF
--- a/docs/content/step.md
+++ b/docs/content/step.md
@@ -393,6 +393,14 @@ After `.worktreeinclude` selects entries, you can add more gitignore-style exclu
 exclude = [".cache/", ".turbo/"]
 ```
 
+To copy nothing unless `.worktreeinclude` exists — matching Claude Code desktop, where the file is required — pass `--require-include`:
+
+```bash
+wt step copy-ignored --require-include
+```
+
+Without `.worktreeinclude`, the command is a no-op (it reports that nothing was copied and why). With the file present, only matching files copy as above. To apply this across every repository, put the flag in a user-config hook: `post-start = "wt step copy-ignored --require-include"`.
+
 ### Common patterns
 
 | Type | Patterns |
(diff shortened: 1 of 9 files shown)


max-sixty:
thanks @yzx9 

so I understand: is the benefit of this that you might not want to copy lots of files?

...but then why have the hook? is the hook on user configs, which downst want to copy on repos.thats aren't configured with the include file?

yzx9:
@max-sixty Yes, I want to put this hook in my user configuration so it applies to all of my repositories. The issue is that some repositories have very large `.gitignore` files, but also include a `.worktreeinclude` file from Claude Code, so I’d like the hook to use that when available.

Also, I think having a single file determine the behavior of this API is more straightforward than having two files conditionally determine it, and it also aligns better with Claude Code’s option.

In practice, I’m already using a workaround, though I’m not sure whether it’s robust enough:

```sh
if [ -f .worktreeinclude ]; then wt step copy-ignored; fi
```

max-sixty:
wdyt about putting it on a flag to the command?

yzx9:
Yes, I think that should be enough. That was my initial thought too, but I saw that a similar case in the `merge` command added a config, so I followed the same approach. I can remove it if you think it’s unnecessary.

max-sixty:
yes I think let's have it as a flag, then we keep the config smaller. @worktrunk-bot could you draft this here (assuming @yzx9 is OK with it), and come up with ideas for what the flag should be called?

yzx9:
Thanks for your work. I squashed the two commits into one for a cleaner history.

max-sixty:
super, thanks @yzx9 !

PR #3204, opened 2026-06-24
fix(switch): match encoded Azure project paths

### LINKED ISSUE #3203: Azure DevOps pr: switch fails when project name contains spaces
## Description

`wt switch pr:<number>` can fail for Azure DevOps repos when the Azure project name contains spaces.

Azure returns decoded project names from `az repos pr show`, for example `project with spaces`, while Git remote URLs store path segments URL-encoded, for example `project%20with%20spaces`. Worktrunk compares those strings directly when finding the matching remote, so it can reject the correct remote.

Observed output shape:

```text
◎ Fetching PR #101...
  fix(example): demo PR title (#101)
  by @alice@example.com · active · fix/example-branch · https://dev.azure.com/example-org/project with spaces/_git/example-repo/pullrequest/101
✗ No remote found for example-org/example-repo
↳ Add the remote: git remote add upstream https://dev.azure.com/example-org/project with spaces/_git/example-repo
```

## Expected

Worktrunk should treat decoded and URL-encoded Azure path segments as the same remote path, then switch to the PR branch.

## Context

A fix is forthcoming with a regression test covering an encoded Azure project path and decoded Azure API project name.

jonasherfort:
Fixes #3203.

Azure DevOps returns decoded project names from `az repos pr show`, while Git remotes store project path segments URL-encoded. Canonicalize Azure path segments before comparing remotes and constructing Azure URLs, so `wt switch pr:N` works for projects like `project with spaces`.

Added an integration regression covering an encoded Azure project path with a decoded Azure API project name.

### DIFF
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -81,6 +81,7 @@ pub use repository::{
 };
 pub use url::parse_owner_repo;
 pub use url::{GitRemoteUrl, GitRepoInfo, GitRepoProvider};
+pub(crate) use url::{canonical_url_path_segment, url_path_segments_eq};
 /// Why branch content is considered integrated into the target branch.
 ///
 /// Used by both `wt list` (for status symbols) and `wt remove` (for messages).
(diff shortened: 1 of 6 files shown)


max-sixty:
thanks @jonasherfort

PR #3209, opened 2026-06-24
docs: design proposal for more wt switch picker shortcuts (built-in + custom template-bound)

max-sixty:
Design proposal only — no production code. The full document is `design/picker-shortcuts.md`; this summarizes the recommendation and the decisions a reviewer needs to weigh.

## What it answers

Two questions about the `wt switch` interactive picker: (1) which built-in keyboard shortcuts are worth adding, and (2) whether to let users bind a key to a template that expands against the selected row, the way `[aliases]` and hooks already do.

## Recommendations

**Built-in shortcuts.** `ctrl-r` to reload the worktree list (pure worktrunk, no OS dependency, and it exercises the same `reload`-verb plumbing custom shortcuts reuse). `alt-y` (copy branch) and `ctrl-o` (open the row's PR/MR URL) once a small cross-platform clipboard/open helper exists — worktrunk has none today, and that helper is the real cost, not the binds.

**Custom shortcuts.** Worth building, tightly scoped: user-config only, under `[switch.picker.shortcuts]`, one key to one template, expanded against the selected row and run in the background while the picker stays open. They fill a gap `[aliases]` can't: side-effects on a worktree you haven't switched to (copy, open, kick a build), since aliases always run against the current worktree.

**Execution.** Reuse the `alt-r` machinery — a new `run` verb on the existing `PickerCollector`, expanding via `expand_template(..., Posix, ...)` and spawning on a background thread, with PR #3199's cursor reposition. An `execute-silent` plus a hidden `wt` helper is the lighter-touch alternative; the doc weighs both.

## Decisions a reviewer should weigh

- **User-config only is a scope choice, not a security necessity.** A project shortcut could route through the same read-only approval the picker already uses for removal hooks, but it would run only if its exact command were approved at a prior gate — which a picker-only command almost never is — so it would usually be an inert key the user can't fix mid-render. User-only drops that footgun and matches where keybindings already live (`[switch.picker]`, `worktree-path`, `[list]` custom-columns are all user-only). Because the picker never loads project shortcuts, the "Project Commands Run Only After Approval" invariant holds structurally.
- **Route (A) reload vs route (B) execute-silent helper.** (A) keeps one expansion site with full live row context but pays a `reload` round-trip and a brief cursor flash per key-press; (B) avoids the reload but pays a subprocess per press and token-only context. (A) is recommended; the cursor flash should be confirmed on a real picker before committing. This is the one call most likely to flip.
- **Semantics: run-and-stay, side-effects only.** Switching stays on Enter; worktree-set mutations leave the list stale (refreshed by `ctrl-r`) rather than being guarded, consistent with how `[aliases]` runs user commands unguarded.

## Source grounding

The doc cites the picker bind block, `PickerCollector`, the approval model (`HookSource`, `approve_readonly`), the template engine (`expand_template`, `validate_template`, `ValidationScope`), the row model, and `SwitchPickerConfig`, with an implementation map in the appendix. The technical claims and the design reasoning were each run through an adversarial review pass against the source; corrections from it are folded in (notably reframing project-config shortcuts as feasible-but-inert rather than impossible).

No code is changed, so there is nothing to test; the doc lives in `design/` to stay outside the `docs/content` sync test.


max-sixty:
Closing as superseded. #3233 landed the built-in half a day after this was opened — `alt-y` (copy branch), `alt-o` (open PR/MR), `alt-r` (refresh) and the cross-platform clipboard/open helper now at `src/commands/picker/os.rs`. That helper was this doc's stated main cost ("worktrunk has none today"), so the premise no longer holds, and #3268 has since dealt with the cursor flash the doc flagged as the thing to confirm before committing.

The doc's recommended execution route has also been overtaken: it argued for reusing the `reload` machinery, and #3233 deliberately went the other way for the built-ins — native skim `Action::Custom` callbacks, so the cursor doesn't reset and `--prs` rows aren't dropped.

What remains live is the custom template-bound shortcuts question — `[switch.picker.shortcuts]`, which is still unbuilt (`SwitchPickerConfig` carries only `pager`). That's better raised fresh against the native-callback machinery that now exists than kept alive here under stale grounding.

PR #3215, opened 2026-06-24
fix(config): honor CLAUDE_CONFIG_DIR for Claude Code paths

tftio:
## The bug

Claude Code reads its configuration from the directory named by the `CLAUDE_CONFIG_DIR` environment variable when it is set, falling back to `~/.claude` otherwise. The override relocates the **entire** config tree, so settings live at `$CLAUDE_CONFIG_DIR/settings.json` and plugins at `$CLAUDE_CONFIG_DIR/plugins/...` (not under a nested `.claude/`). This is the documented mechanism for running multiple accounts side by side, e.g. `alias claude-work='CLAUDE_CONFIG_DIR=~/.claude-work claude'`.

worktrunk hardcoded `~/.claude` in every Claude Code path and never consulted `CLAUDE_CONFIG_DIR`. On a machine that uses the override:

- `wt config show` reports the worktrunk plugin and the statusline as **not** installed even when they are — it reads `~/.claude/plugins/installed_plugins.json` and `~/.claude/settings.json`, which Claude Code isn't using.
- `wt config plugins claude install-statusline` writes to the **wrong** file: it creates/edits `~/.claude/settings.json`, silently materializing a stray directory while never touching the `settings.json` Claude Code actually loads. The statusline appears to install successfully but never shows up.

The affected call sites were all in `src/commands/config/`:

| Location | Symptom |
|----------|---------|
| `show.rs` `is_plugin_installed()` | plugin reported uninstalled |
| `show.rs` `is_statusline_configured()` | statusline reported unconfigured |
| `plugins.rs` `require_settings_path()` | statusline written to the wrong file |

## The fix

Add a single `claude_config_dir()` helper next to the existing `home_dir()` helper. It returns `$CLAUDE_CONFIG_DIR` when set (expanding a leading `~/` against the home directory, since a literal `~` only reaches us when the variable is set outside a shell), and falls back to `~/.claude` otherwise. The three call sites above now resolve their paths through it, and the `install-statusline` help text / directory-creation message no longer hardcode `.claude`.

The Gemini extension path (`~/.gemini/...`) is a separate tool and is intentionally untouched.

## Tests

- New integration test `test_plugins_claude_install_statusline_honors_claude_config_dir`: with `CLAUDE_CONFIG_DIR` pointed at a custom directory, `install-statusline` writes there (and **not** to `~/.claude`), and a second run detects the relocated statusline as already configured.
- `set_temp_home_env` now also pins `CLAUDE_CONFIG_DIR` to the temp home's `.claude`, mirroring the existing `OPENCODE_CONFIG_DIR` handling. This keeps the Claude Code detection tests hermetic against an ambient `CLAUDE_CONFIG_DIR` inherited from the runner's environment — without it, the suite is non-deterministic on any machine (such as a Claude Code session) that sets the variable.

Locally verified: `cargo fmt --check`, `cargo clippy --all-targets`, the full `config_show` module (covers every detection/install path), and the full `user_hooks` module (covers hook env behavior, the only place the `set_temp_home_env` change could ripple) all pass.


max-sixty:
thanks @tftio !

max-sixty:
@worktrunk-bot if you can push coverage for those then great

PR #3225, opened 2026-06-25
fix(switch): repaint the picker preview after the alt-r sticky reposition

max-sixty:
## What

Two linked fixes around the alt-r "remove worktree" path in the interactive picker:

1. **Picker (behavior):** the preview pane now repaints after the alt-r sticky reposition, so it tracks the cursor.
2. **Test (robustness):** `test_switch_picker_alt_r_keeps_cursor_sticky` no longer races the preview pane.

## The race

When alt-r removes a worktree, skim resets the cursor to the top on reload, and an injected `Custom` action repositions it onto the row that slides into the removed slot — that part works. But skim only re-runs the preview on a selection-*change* event (`on_selection_changed` → `Event::RunPreview`). Moving the cursor through the `ItemList` API inside that `Custom` action is not such an event, so skim kept showing the row it last previewed — the current worktree, which the reload had briefly reset the cursor to. The preview pane went stale until the next keystroke.

The flaky test asserted cursor position by reading that preview pane, so under load it timed out (~10%) waiting for `"<sticky-row> has no uncommitted changes"` text that the stale pane never showed — even though the cursor had landed correctly. Captured failure (note the `>` pointer is on `wt-keep`, but the pane shows `main`):

```
  @ main         ^                  │
> + wt-keep      _                  │○ main has no uncommitted changes
```

## The fix

- `reposition_cursor_action` returns `Event::RunPreview` once it lands the cursor, so the preview repaints for the sticky row — the same idiom the preview-tab keybindings already use. This fixes the real (if minor) UX papercut where the preview lagged the cursor after a removal.
- `drive_alt_r_then_switch` gates on the **list-pane cursor pointer** (`> ` on the target row) rather than the preview pane. The pointer is drawn on every item-list render, so it is a race-free signal of cursor position — which is exactly what this test verifies. New `wait_for_cursor_on_row` helper, built on a refactored predicate-based `wait_for_stable_until` that `wait_for_stable_with_content` now also delegates to.

The two fixes are independent: the test gates on the cursor, not the preview, so it is robust whether or not the picker repaints.

## Sibling scan

`switch_picker.rs` is the only file with interactive picker screen-waits. The preview-panel and tab-cycle tests gate on preview content reached by **key navigation** (Down / alt-N / Tab), which *does* fire `RunPreview`, and the preview content is their actual subject — they don't share this flaw. The query-based proxy tests (`emits_cd_directive`, `no_cd`, `pre_switch_hook`) have a separate documented matcher-lag rationale. Only the alt-r `Custom`-action reposition had no preview refresh; the picker fix structurally protects any future alt-r-then-preview observation.

## Validation

Reproduced the flake under load (~3/20 fail before). After the fix:

- Test fix alone (no picker change): robust across many runs under heavy load.
- Picker fix verified causally — with a strict preview-content gate the preview reliably tracks the cursor with the fix, and reverting *only* the picker fix reintroduces preview-gate timeouts. So the picker fix is what makes the preview follow the cursor.
- Full gate (`cargo run -- hook pre-merge --yes`, includes `--features shell-integration-tests`): 4187 passed, all lints green, no snapshot changes.

### DIFF
--- a/src/commands/picker/mod.rs
+++ b/src/commands/picker/mod.rs
@@ -422,6 +422,13 @@ fn sticky_reposition_target(removed_pos: usize, remaining_data_rows: usize) -> O
 /// from [`sticky_reposition_target`]. Under an active fuzzy query the displayed
 /// order diverges from `shared_items` order, so the landing row is approximate
 /// (a valid nearby row) rather than the exact next row.
+///
+/// On landing, it returns [`Event::RunPreview`] so the preview pane repaints for
+/// the row the cursor settled on. skim only repaints the preview on a
+/// selection-*change* event (`on_selection_changed`), and moving the cursor
+/// through the `ItemList` API here is not one — without this, the pane keeps
+/// showing the row skim last previewed (the current worktree, which the reload
+/// briefly reset the cursor to) until the next keystroke.
 fn reposition_cursor_action(
     target: usize,
     attempts: Arc<AtomicUsize>,
@@ -429,12 +436,13 @@ fn reposition_cursor_action(
 ) -> Action {
     Action::Custom(ActionCallback::new_sync(
         move |app| -> Result<Vec<Event>, Box<dyn std::error::Error + Send + Sync>> {
-            // Rows are in: land the cursor on the removed row's slot.
+            // Rows are in: land the cursor on the removed row's slot, then
+            // repaint the preview for it (the cursor move alone doesn't).
             if app.item_list.count() > 0 {
                 app.item_list.jump_to_first();
                 app.item_list
                     .scroll_by(i32::try_from(target).unwrap_or(i32::MAX));
-                return Ok(Vec::new());
+                return Ok(vec![Event::RunPreview]);
             }
             // No rows yet. The matcher has "settled" once it has stopped with the
             // reloaded items taken and a non-empty pool (the empty pool is the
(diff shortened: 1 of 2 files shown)


max-sixty:
### Note on the red `codecov/patch`

The one uncovered patch line is the picker fix itself:

```rust
return Ok(vec![Event::RunPreview]);   // the alt-r reposition "landing" branch
```

This is a **pre-existing, structural coverage gap, not a regression**. On `main` the entire reposition landing branch (`count() > 0`, `jump_to_first`, `scroll_by`, and the prior `Ok(Vec::new())`) is already uncovered (`base=0` in the codecov compare API) — the branch runs inside skim's tokio event-loop `Custom` callback, which the PTY test exercises but LLVM coverage doesn't capture in the `wt` subprocess (the instrumented binary doesn't flush counters before the test reclaims the child). Editing the line simply moved it into the patch denominator.

It can't be covered by an in-process unit test either: skim's `ActionCallback::call` is `pub(crate)`, and `App` / `item_list` can't be populated from outside the crate, so the callback body isn't drivable from a test.

The branch is exercised end-to-end by `test_switch_picker_alt_r_keeps_cursor_sticky` (which now gates on the cursor pointer and is green on all three `test` platforms). Merging with this `codecov/patch` miss accepted.

PR #3264, opened 2026-06-27
test(picker): de-flake preview-auto-refresh by filtering to the PR row

max-sixty:
The `ci` workflow has been failing on `main`: `test_switch_picker_preview_auto_refreshes_when_compute_lands` times out (30s) on every platform, jumping between linux/macos/windows across runs. It is a genuine flake, not a regression from a specific commit. The failing `code-coverage` job shares the same single root cause — it runs the suite under llvm-cov.

Root cause: the test drives a `wt switch --prs` picker with two rows — the `main` worktree row and a not-checked-out PR row (`flaky`) — then has to land the cursor on the PR row before opening its comments tab. Two things make that fragile. The order between the rows isn't fixed (the `--prs` row streams in on its own thread, and the worktree row's `PathName` tiebreak reads the random temp path, so either order can win), and skim's cursor resets to the top on each async refresh — so a single Down lands on whichever row happens to sit below the cursor. When it landed on the worktree row, the test waited forever for the PR row's "not checked out" preview.

Fix: isolate the PR row with skim's inverse-match operator `!main`, which excludes the worktree row by the one token only it carries — its branch name — leaving the `--prs` row as the sole, selected row. The cursor is then on it deterministically, regardless of row order, and the gate matches the preview text directly rather than depending on a cursor/preview screen-line coincidence.

(A first attempt filtered with `#`, the PR gutter sigil, but the mock returns this PR for every `gh pr list` — including the worktree row's CI-status lookup — so the PR's number/title/author fold into the worktree row's matcher text too, and no PR-content query can single out the PR row. The worktree branch name is the one token it doesn't share.)

Verified: deterministic single-row isolation across repeated runs, and 20× green under heavy CPU oversubscription (the flake reproduces only under contention). The full `switch_picker` module (37 tests) is green.

### DIFF
--- a/tests/integration_tests/switch_picker.rs
+++ b/tests/integration_tests/switch_picker.rs
@@ -1475,8 +1475,8 @@ fn test_switch_picker_preview_auto_refreshes_when_compute_lands(mut repo: TestRe
     // cache hit. `pr list` is instant so the row lands promptly.
     let mock_bin = repo.root_path().join("mock-bin");
     std::fs::create_dir_all(&mock_bin).unwrap();
-    // A short head branch so it isn't truncated in the narrow (preview-shown)
-    // list pane — the test gates on its full text to confirm the row rendered.
+    // A short head branch so the PR row isn't truncated in the narrow
+    // (preview-shown) list pane.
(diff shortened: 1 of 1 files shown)


max-sixty:
@worktrunk-bot lets set these flakey tests to be ignored, and add an issue to restore them

max-sixty:
also ignore anything similar that might flake + add to the issue

PR #3265, opened 2026-06-27
test(picker): de-flake preview-auto-refresh by filtering to the PR row

max-sixty:
`test_switch_picker_preview_auto_refreshes_when_compute_lands` has been timing out on every platform across runs (the most recent: it failed the advisory affected-tests jobs on linux and windows while passing the required full-suite runs — a classic flake).

**Root cause:** a `--prs` row's order against the worktree row isn't fixed — it streams in on its own thread, and an async refresh resets skim's cursor to the top, so the order can settle as either `[worktree, pr]` or `[pr, worktree]`. The test pressed a single `Down`, assuming `[worktree, pr]`. When the order was `[pr, worktree]`, the cursor reset landed on the PR row and `Down` moved *off* it onto the worktree row, where the Down-nudge then clamped — stranding the test for the full 30s.

**Fix:** filter to the PR row with `#` (its gutter sigil; a worktree row's path can't contain a literal `#`, so this isolates PR/MR rows). The PR row becomes the sole, selected row, removing the navigation entirely and the order dependency with it. The gate now matches the preview text anywhere on screen rather than requiring a cursor/preview screen-line coincidence.

Verified locally: passes on repeated back-to-back runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,75 @@
 # Changelog
 
+## 0.62.0
+
+### Improved
+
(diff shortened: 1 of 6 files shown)


max-sixty:
@worktrunk-bot merge main, finished this off

PR #3356, opened 2026-07-03
docs(skills): make /wt-switch-create worktree creation unconditional

max-sixty:
`/wt-switch-create` too often skipped creating the worktree, judging that a research or read-only task didn't need isolation. Two wordings invited that: the lead was only an ordering rule ("steps 1–3 come before any other work"), and Scope's "authorizes" read as permission, prompting a should-I test that user-level rules gating worktree creation behind explicit request answer with "no".

The lead now states that invoking the command is itself the explicit request (a research task gets a worktree all the same), Scope states the mandate's bounds without permission language, and rationale.md records the failure mode so the wording isn't softened back.

Prose-only; no code or generated docs touched.

### DIFF
--- a/skills/wt-switch-create/SKILL.md
+++ b/skills/wt-switch-create/SKILL.md
@@ -33,7 +33,9 @@ branch-shaped lead — all task).
 
 ## What to do
 
-Steps 1–3 come before any other work.
+Steps 1–3 run on every invocation, before any other work. The invocation is
+itself the explicit request to create the worktree; a research or read-only
+task gets one all the same.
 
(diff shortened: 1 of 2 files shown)


cedoi93-svg (approved):

cedoi93-svg (approved):

cedoi93-svg (approved):

cedoi93-svg (approved):
Leave a review

cedoi93-svg (approved):

cedoi93-svg (changes requested):
Review

PR #3453, opened 2026-07-13
fix(plugin): anchor WorktreeCreate at the project dir

silvanshade:
**Anchor the `WorktreeCreate` hook at `CLAUDE_PROJECT_DIR`.** The hook command inherits the session shell's working directory. When the shell has `cd`'d outside the repository (e.g. into a reference-papers directory elsewhere on disk), `wt switch` dies with *not a git repository* and every isolated-agent spawn in the session fails — observed as a six-agent spawn failure mid-session. `cd "${CLAUDE_PROJECT_DIR:-.}"` pins the hook to the repository the session was launched in, falling back to the current behavior when the variable is unset.

_Originally this PR also added `--base=@` to base agent worktrees on the session's HEAD. That change was dropped at the maintainer's request (worktrunk-bot pushed the removal) so it can be evaluated separately — hardcoding `--base=@` would override the native `worktree.baseRef` setting for everyone, and the hook has no way to read that setting. This PR is now the `cd` anchor only._

### DIFF
--- a/plugins/worktrunk/hooks/hooks.json
+++ b/plugins/worktrunk/hooks/hooks.json
@@ -69,7 +69,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'name=$(jq -er .name) || exit 1; bash \"$CLAUDE_PLUGIN_ROOT/hooks/wt.sh\" switch --create \"$name\" --no-cd --format=json | jq -er .path'"
+            "command": "bash -c 'name=$(jq -er .name) || exit 1; cd \"${CLAUDE_PROJECT_DIR:-.}\" || exit 1; bash \"$CLAUDE_PLUGIN_ROOT/hooks/wt.sh\" switch --create \"$name\" --base=@ --no-cd --format=json | jq -er .path'"
           }
         ]
       }
--- a/plugins/worktrunk/hooks/hooks.json
+++ b/plugins/worktrunk/hooks/hooks.json
@@ -69,7 +69,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'name=$(jq -er .name) || exit 1; cd \"${CLAUDE_PROJECT_DIR:-.}\" || exit 1; bash \"$CLAUDE_PLUGIN_ROOT/hooks/wt.sh\" switch --create \"$name\" --base=@ --no-cd --format=json | jq -er .path'"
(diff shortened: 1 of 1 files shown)


max-sixty:
> One thing I can't approve on my own: this file (`plugins/worktrunk/hooks/hooks.json`) is on worktrunk's data-loss surface — it also ships the `WorktreeRemove` hook that runs `wt remove` — so per the repo's review policy any edit here goes to a human rather than a bot approval. Routing to @max-sixty for the merge call. The diff doesn't touch the removal path, so this is a policy hold, not a concern about the change.

just as a process point, this is overly conservative, @worktrunk-bot you should be deciding based on whether the code actually touches code that is in data-destruction paths, not whether there are unrelated changes in the file; please fix the guidance here

max-sixty:
thanks @silvanshade 

Have you run with this at all yourself? I can test it myself for a while. I think your points all make sense; particularly (1); but I'm nervous about changes to this area of code given they're difficult to test.

@worktrunk-bot could you compare the default behavior of Claude Code to these?

silvanshade:
> Have you run with this at all yourself? I can test it myself for a while. I think your points all make sense; particularly (1); but I'm nervous about changes to this area of code given they're difficult to test.

Yes, I have been running with this configuration since Fable identified it as a fix to workflow problems we were having.

The scenario described is real, I have a few separate repos which are topically related but not formally connected via submodules or anything like that, and then additional directories containing relevant research artifacts.

Prior to the change, worker subagents would sometimes get tripped up during tasks that would involve navigating around to do some analysis and then interacting with worktrunk.

Since the change, reports about those issues from the coordinator agents have stopped. In that regard it seems to have addressed the problem. Whether it's actually the right solution here, I don't know.

max-sixty:
> but kept head as a first-class supported mode

how is this mode selected?

max-sixty:
happy to merge the `cd` now!

max-sixty:
@worktrunk-bot can you remove the `--base` here so that @silvanshade keeps the PR?

max-sixty:
> It's a settings key, not a flag or a per-run choice — [`worktree.baseRef`](https://code.claude.com/docs/en/worktrees#choose-the-base-branch) in `settings.json`, accepting only `"fresh"` (default) or `"head"`:

Do plugins have access to this config?

max-sixty:
OK, @silvanshade — my sense is we should stay with the default, at least until CC allows for reading its config — how do you feel about that?

We could also quite easily add a config for default args for the hook — lmk and I'll implement?

PR #3503, opened 2026-07-17
fix(testing): Isolate CLI logging environment

reneleonhardt:
## 🧢 Changes

- Remove inherited `RUST_LOG` from CLI integration-test children.
- Preserve explicit per-test logging overrides.
- Keep verbose trace assertions independent of the developer's shell.

## ☕️ Reasoning

The test harness should control the child CLI's logging baseline. A host
`RUST_LOG=warn` overrides `-vv`, suppresses debug trace records, and makes
otherwise-correct diagnostic tests fail.

This was a real-world agentic coding blocker: full verification inherited the
session environment and required an external `env -u RUST_LOG` workaround.

## 🧪 Reproduction

```sh
RUST_LOG=warn cargo test --test integration \
  test_diagnostic_trace_log_contains_git_commands -- --exact
```

Before this fix, the assertion failed because `trace.log` lacked the expected
Git command records:

```text
Trace log should contain git worktree list command
```

## ✅ Verification

- Focused harness test proves `RUST_LOG` is explicitly removed.
- The diagnostic regression passes with host `RUST_LOG=warn`.
- Combined publication-set pre-merge hook: 4,519 tests passed on macOS.

Developed with carefully directed, manually reviewed AI assistance.

Co-Authored-By: GPT-5.6 Sol <codex@openai.com>

### DIFF
--- a/src/testing/mod.rs
+++ b/src/testing/mod.rs
@@ -546,12 +546,13 @@ pub fn configure_cli_command(cmd: &mut Command) {
     // by the env-strip above.
     isolate_subprocess_env(cmd, None);
     cmd.env("WORKTRUNK_TEST_EPOCH", TEST_EPOCH.to_string());
-    // RUST_LOG intentionally NOT set: the flag baseline and `RUST_LOG`
+    // Do not inherit the host's RUST_LOG: the flag baseline and `RUST_LOG`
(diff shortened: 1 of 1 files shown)


max-sixty:
this one I'll take @reneleonhardt ! thanks

PR #3662, opened 2026-07-30
Simplify forge, shell integration, and removal internals

max-sixty:
This consolidates cross-cutting models that had accumulated parallel representations, while preserving the CLI and config interfaces.

## What changed

- Forge identity now flows through one `ForgeKind`, with boundary-safe network-host classification shared by CI, remote references, and structured repository metadata. Branded SSH aliases such as `github-personal` remain outside provider dispatch and receive an actionable `forge.platform` diagnostic when forge data is requested.
- Worktree removal now uses one owned target type and rechecks uncached worktree topology immediately before compare-and-swap branch deletion, retaining branches that gained a live or locked checkout.
- Shell integration no longer implements the retired single-file directive writer. Stale wrappers receive repair guidance, execution fails closed, and child processes cannot inherit the retired sourceable-file capability.
- Zsh and Git-version probes are shared, while dead mocks, redundant dependency edges, serializer detours, pass-through types, and duplicate tests are removed.

The removal guard deliberately distinguishes live, stale-prunable, and locked registrations. The detached cleanup path performs the same record-aware check before deleting a branch ref.

## Testing

`cargo run -- hook pre-merge --yes` passed after merging current `origin/main`: 4,489 tests, Clippy, formatting, lockfile checks, docs, doctests, and snapshot review.

> _This was written by Claude Code on behalf of max_.

PR #3673, opened 2026-07-30
fix(forge): classify a forge by the brand in the hostname, not by DNS label

### LINKED ISSUE #3671: Widen the forge diagnostic to hosts that lost exact-label classification
Since 0.71.0, forge hosts are classified by exact DNS label (#3662). That correctly rejects lookalikes like `evil-github.com` and `dev.azure.com.attacker.example`, but it also drops legitimately-named self-hosted instances whose hostname carries the forge name only as part of a label:

- `gitlab-internal.company.com`
- `github-enterprise.acme.com`
- `mygithub.com`

Those lose CI status in `wt list --full`, `wt switch --prs`, and `repo.provider` in `wt list --format=json`. The remedy is one line — `forge.platform` in `.config/wt.toml` — but nothing tells the user to set it.

Only one shape announces itself today. `LegacyForgeAlias::from_url` (`src/git/ci_platform.rs`) requires an SSH URL **and** a dotless host **and** a `github-`/`gitlab-`/`gitea-` prefix, so it catches the branded multi-account alias (`git@github-personal:owner/repo.git`) and nothing else. A dotted hyphenated host gets no message at all — CI simply stops appearing, which reads as a bug rather than a configuration gap.

## Proposal

Widen the diagnostic to any host that would have matched the pre-0.71.0 substring rule but no longer classifies: if `forge.platform` is unset, the host has no exact forge label, and a case-insensitive substring of the host is `github`/`gitlab`/`gitea`, emit the existing `forge.platform` hint with the inferred platform.

That is additive — it changes no classification and needs no further breaking change — and it keeps dispatch on the exact-label boundary, which is the security property #3662 bought. The `LegacyForgeAlias` doc comment already states the rule that matters here: the type is diagnostic-only and must not be used to select a provider.

Worth checking whether the same widening should cover Azure DevOps, where the equivalent case is a host ending in neither `dev.azure.com` nor `visualstudio.com` but containing one of them.

## Context

Found during the 0.71.0 release review. The behavior is deliberate and documented in the changelog with the remedy; this issue is only about making the affected user aware without having to read release notes.

max-sixty:
Reverts the branded half of the exact-label classifier and deletes the
diagnostic built to explain it. `github-enterprise.acme.com`, `mygithub.com`,
`gitlab-internal.company.com`, and the `github-personal` SSH alias resolve to
their forge again, so CI status, `wt switch --prs`, and `repo.provider` work
with no config.

## Why the label boundary goes

It looked like an ownership check and wasn't one. An attacker controls their own
DNS, so `github.attacker.example` has the exact label `github` and classified
fine; `gitlab.evil.co.uk` likewise. What the rule actually excluded was the
self-hoster who put the brand in a hyphenated name. It failed open for the
adversary and closed for the customer.

The residual case for it doesn't survive either. The hostname comes out of the
user's own `.git/config`, and whoever can put a host there can put code there
too — the trust decision happens at clone time, and by the time worktrunk reads
the remote the user is already building from it. All the classification decides
is which forge CLI (`gh`, `glab`, `tea`, `az`) runs against it.

So the rule is recall-first: any host carrying `github`, `gitlab`, or `gitea`
matches, first match winning. The cost is a host that merely sounds like a forge
getting a forge CLI run at it, which surfaces as that CLI's error rather than as
silence — the better of the two failures, and `forge.platform` overrides it.

## What stays

Azure DevOps keeps suffix matching on its two service domains, and for a reason
unrelated to security: those are service domains rather than a brand in the
host, so every real hosted instance already matches, and the on-prem edition
carries neither string. `dev.azure.com.attacker.example` and
`evil-visualstudio.com` are outside the domains and carry no brand to fall back
on, so they stay unclassified. Userinfo still resolves to the network host, so
`https://github.com@attacker.example/…` is `attacker.example`.

## What goes

`LegacyForgeAlias`, `Repository::legacy_forge_alias`,
`legacy_forge_alias_diagnostic`, and its three emit sites in `wt list`,
`wt switch --prs`, and `wt config show --full`. Every host the diagnostic fired
on now classifies, so it could only ever return `None`. A host with no brand at
all still reaches the existing generic hint, which is the right message there —
there is no platform to infer.

The end-to-end warning-dedup test goes with it, since no warning is raised from
both the collect and `--prs` threads any more; `stash_warning_preserves_order`
keeps the mechanism covered.

## Docs

`## Forge platform` in `src/cli/mod.rs` described the override as being for SSH
aliases and self-hosted instances, which now detect on their own. It states the
rule and scopes the override to hosts carrying no brand — a Forgejo instance at
`forge.example.com`. Mirrors, `dev/wt.example.toml`, and the two `config` help
snapshots regenerate from it.

---

The branch's history has a false start — the first commit widened the diagnostic, the second deletes it in favour of relaxing classification — plus a merge of `main` after v0.71.0 shipped. The net diff is the second approach; it all squashes on merge.

Also supersedes [#3672](https://github.com/max-sixty/worktrunk/pull/3672), the triage bot's PR for the same issue — it widens the diagnostic rather than removing the need for one, so it should be closed too.

Closes #3671


max-sixty:
Right that v0.71.0 shipped — it's tagged at `aac20ee`, and the exact-label entry is in the released changelog. I've merged `main` in, so this now sits on top of the release.

Not adding a `## 0.72.0` section, though. Entries here are written when the version is cut, not by the PRs that go into it: `## 0.71.0` was still absent from `CHANGELOG.md` while #3662, #3664, #3665, #3666, and #3669 sat merged on `main`, and the `release` skill puts "Add `## X.Y.Z` section at top with changes" at step 9 of the release itself. Step 12 is the backstop for exactly this case — it lists everything that reached `main` since the cut and folds what's user-facing into the changelog before the tag, so the reversal gets documented when 0.72.0 is cut, alongside whatever else lands first.

The concern behind the suggestion is right and worth carrying forward: a user who read 0.71.0's breaking-change note should see it rolled back, so the 0.72.0 entry needs to name the reversal rather than describe the new rule from scratch.

PR #3682, opened 2026-07-31
docs: add MiniMax commit generation example

octo-patch:
Reason: Add a documented MiniMax option for Worktrunk commit message generation.

## Summary

- Document a MiniMax-M3 command using the supported global endpoint.
- Note the regional endpoint used in China.
- Refresh generated config references, skill mirrors, and help snapshots.

## Checks

- `cargo fmt --check`
- `cargo test --test integration readme_sync`
- `cargo test --test integration 'help::test_help'`

### DIFF
--- a/dev/config.example.toml
+++ b/dev/config.example.toml
@@ -64,6 +64,13 @@
 # [commit.generation]
 # command = "MAX_THINKING_TOKENS=0 claude -p --no-session-persistence --model=haiku --tools='' --safe-mode --setting-sources='user' --system-prompt=''"
 #
+# ### MiniMax
+#
+# [commit.generation]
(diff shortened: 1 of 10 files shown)


max-sixty:
Thanks for putting this together. Closing it though: the examples are meant to cover distinct harnesses rather than each provider reachable through one that's already documented.

This is the same `claude -p` command with `ANTHROPIC_BASE_URL` and `ANTHROPIC_MODEL` pointed elsewhere, so it shows a reader nothing the Claude example doesn't. Any Anthropic-compatible endpoint works that way, and adding them one at a time grows the identical command across the doc and skill mirrors plus help snapshots.

Codex earns its entry by being a different CLI with its own flags. An example for a harness that isn't covered yet would be welcome.

Issue #2643, opened 2026-05-08
Bug Report: `wt switch` fails with `zoxide: no match found` when `cd` is aliased to `__zoxide_z`


xkumiyu:
### Summary

When [zoxide](https://github.com/ajeetdsouza/zoxide) is initialized in the shell **before** worktrunk's shell integration, `wt switch` fails to change directories and prints `zoxide: no match found`. This happens because `eval "$(command wt config shell init zsh)"` is evaluated after zoxide has already aliased `cd` to `__zoxide_z`. As a result, the `cd -- "$(<"$cd_file")"` line in the generated `wt()` function is stored internally as `__zoxide_z -- "$(<"$cd_file")"`. Since `__zoxide_z` queries zoxide's internal database rather than accepting arbitrary absolute paths, the directory change fails.

### Environment

- **OS**: macOS / Linux (zsh)
- **Shell**: zsh with zoxide initialized via `eval "$(zoxide init zsh)"` **before** worktrunk shell integration
- **Relevant zoxide alias**: `cd` → `__zoxide_z`

### Steps to Reproduce

1. Initialize zoxide in `.zshrc` so that `cd` is aliased to `__zoxide_z`
2. Afterwards, initialize worktrunk shell integration: `eval "$(command wt config shell init zsh)"`
3. Run `wt switch <worktree-name>`

### Actual Behavior

```
○ Switched to worktree for ...
zoxide: no match found
```

The worktree switch is reported as successful, but the shell does not change to the target directory. This occurs even for directories that have been previously visited (i.e., already in zoxide's database).

### Expected Behavior

The shell changes to the worktree directory regardless of whether `cd` has been aliased by a third-party tool.

### Root Cause

In the shell integration script (`wt config shell init zsh`), the `wt()` function uses `cd` to navigate to the path written to the directive file:

```bash
if [[ -s "$cd_file" ]]; then
    cd -- "$(<"$cd_file")"
    ...
fi
```

If zoxide is already active at `eval` time, `cd` in this function body is resolved to `__zoxide_z`. This can be confirmed by inspecting the evaluated function:

```zsh
❯ print -r -- "${functions[wt]}" | grep "cd"
        __zoxide_z -- "$(<"$cd_file")"
```

`__zoxide_z` does not accept arbitrary absolute paths — it queries zoxide's scored database, so the directory change fails.

Using `builtin cd` in the source instead of bare `cd` would bypass any aliases and resolve this issue.

### Suggested Fix

[...]

PR #3714, opened 2026-08-02
fix(shell): use command rm in wrapper cleanup to bypass rm aliases

Ar4l:
zsh (and bash with `expand_aliases`) expand aliases at function-**definition** parse time. If a user has `alias rm='rm -v'` defined before `eval "$(wt config shell init zsh)"` in their rc file — a common ordering, since alias blocks usually come first — the wrapper's cleanup bakes in `rm -v -f …`, and every `wt` invocation prints two lines of noise:

```
removed '/tmp/tmp.XXXXXXXXXX'
removed '/tmp/tmp.YYYYYYYYYY'
```

This changes the zsh and bash wrapper cleanup from `rm -f` to `command rm -f`, mirroring:

- the existing `command rm -f` in `templates/fish.fish`
- the `builtin cd` guard added for the same alias-expansion reason in the zsh template (#2643)

nushell and powershell are unaffected (`rm` there resolves to shell builtins with different alias semantics).

Insta snapshots regenerated; `cargo test --lib shell` passes (236 tests).

### DIFF
--- a/src/shell/snapshots/worktrunk__shell__tests__init_bash.snap
+++ b/src/shell/snapshots/worktrunk__shell__tests__init_bash.snap
@@ -60,7 +60,7 @@ if command -v wt >/dev/null 2>&1 || [[ -n "${WORKTRUNK_BIN:-}" ]]; then
             fi
         fi
 
-        rm -f "$cd_file" "$exec_file"
+        command rm -f "$cd_file" "$exec_file"
         return "$exit_code"
     }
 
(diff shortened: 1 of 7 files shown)


Ar4l:
Done in 7c803a6 — regenerated the two integration snapshots (`cargo insta`-style accept run of `--test integration --features shell-integration-tests integration_tests::init::test_init`; all 10 cases pass).

Note I trimmed the accept run's rewrite of the `info:` env metadata out of the committed snapshots: my local run isn't under `cargo llvm-cov nextest`, so accepting verbatim would have churned the recorded env block (`LLVM_PROFILE_FILE`, `GIT_CONFIG_*`, etc.) beyond the intended change. The committed diff is exactly `rm -f` → `command rm -f` in each file, and the tests pass against them since insta compares content, not metadata.

PR #3744, opened 2026-08-05
fix(copy): skip source files that vanish mid-copy instead of aborting

### LINKED ISSUE #3743: step copy-ignored aborts entire copy when a source file vanishes mid-walk
## What happened

\`wt step copy-ignored\` hard-fails and aborts the whole copy if a source file disappears between the directory walk (Phase 1) and the parallel copy (Phase 2) in \`copy_dir_recursive\` ([src/copy.rs](https://github.com/max-sixty/worktrunk/blob/main/src/copy.rs)).

This is common when copying a \`target/\` directory that a concurrent \`cargo build\` in the source worktree is actively rewriting — an intermediate object file present during the walk gets deleted before its leaf is copied:

\`\`\`
✗ copying directory target
  reading metadata for /path/to/fs/target/debug/deps/graphql_client_codegen-....rcgu.o
  No such file or directory (os error 2)
\`\`\`

The error surfaces from \`symlink_metadata()\` in \`copy_leaf\` ([src/copy.rs:70-72](https://github.com/max-sixty/worktrunk/blob/main/src/copy.rs#L70-L72)), and because it propagates through rayon's \`try_for_each\`, the entire batch of leaves aborts — even though most of the other files in the same directory copy successfully. Small/static gitignored dirs (\`.venv/lib\`, \`.claude/\`, etc.) are unaffected since they aren't being mutated concurrently, but a large \`target/\` copied from an actively-building worktree fails intermittently and unpredictably.

## Expected behavior

A source file that vanishes between the walk and the copy should be skipped, not treated as a fatal error — mirroring the existing handling for a destination that already exists (\`ErrorKind::AlreadyExists => return Ok(None)\` at [src/copy.rs:102](https://github.com/max-sixty/worktrunk/blob/main/src/copy.rs#L102)).

## Repro

1. Have a large gitignored dir (e.g. a Rust \`target/\`) that's being actively written by a concurrent build.
2. Run \`wt step copy-ignored\` (or let it fire via a \`post-start\` hook) while the build is running.
3. Occasionally the copy aborts partway with a \`No such file or directory\` error on whatever intermediate file the build deleted/replaced during the walk.

Re-running \`wt step copy-ignored\` (safe, since it skips existing files) usually — but not always, if the build is still active — finishes the job. This isn't a reliable fix, since a still-running build can lose the race on every retry.

Draft PR incoming.

dataders:
Fixes #3743.

## Problem

`copy_leaf` propagates `NotFound` from `symlink_metadata`/`read_link`/`reflink_or_copy` as a hard error. `copy_dir_recursive` collects leaves in a walk phase, then copies them in a parallel `try_for_each` — so a single source file that disappears between those two phases (e.g. a concurrent build rewriting `target/`) aborts the *entire* batch, even though most other files copy successfully.

## Fix

Treat `NotFound` on the source the same way the existing `AlreadyExists` case (destination) is already handled: skip the leaf and return `Ok(None)` instead of erroring.

## Testing

- Added `test_copy_leaf_skips_vanished_source` (unit test, deterministic — no race needed).
- `cargo test --lib copy::` — 3 passed.
- `cargo test step_copy_ignored` (full integration suite) — 56 passed.
- `cargo clippy --lib -- -D warnings` — clean.

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -223,10 +223,8 @@ pub fn copy_dir_recursive(
                 // The entry was listed but is gone already — same race.
                 Err(e) if e.kind() == ErrorKind::NotFound => continue,
                 Err(e) => {
-                    return Err(anyhow::Error::from(e).context(format!(
-                        "reading metadata for {}",
-                        entry.path().display()
-                    )));
+                    return Err(anyhow::Error::from(e)
+                        .context(format!("reading metadata for {}", entry.path().display())));
                 }
             };
             let src_path = entry.path();
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -47,6 +47,13 @@ static COPY_POOL: LazyLock<rayon::ThreadPool> = LazyLock::new(|| {
 /// concurrent build deleting/replacing a build artifact). When `force` is
 /// true, existing entries are removed before copying.
 ///
+/// The vanished-source skip never costs the destination a file: the source is
+/// stat'd before `force` removes anything, so a source that is already gone
+/// leaves the destination untouched. A source that disappears in the narrower
+/// window between that stat and the copy is an error under `force` — the
+/// destination has been removed by then, and silently reporting a skip would
+/// leave a hole where a file used to be.
+///
 /// When `root` is `Some`, refuses destination paths whose parent resolves
 /// outside `root`. The check guards the parent chain, not the final leaf, so
 /// leaf-symlink behavior is preserved (without `force` the symlink is skipped;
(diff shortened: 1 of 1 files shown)


max-sixty:
@worktrunk-bot let's make the changes directly and then @dataders can revert if he disagrees

dataders:
thanks @max-sixty (and @worktrunk-bot)!

Been a big fan of your tool! It's been a lifesaver for patching bugs in Fusion/dbt Core 2.0. Saving space across all my Rust worktrees' `target/` dir is crucial on my macbook w/ only 500GB of disk space.

p.s. It's been a minute since I've seen you around in the dbt community! You should take the Rust rewrite (dbt Core 2.0) for a spin when you get a chance.

max-sixty:
ok but I see new comments @worktrunk-bot ?

Issue #3753, opened 2026-08-06
WorktreeRemove hook fails for a path that exists but holds no worktree (skeleton leftover) → agents-view session undeletable


judewang:
### Summary

`WorktreeRemove` short-circuits when the recorded worktree path is gone (#3493), and removes it when a live worktree is there. There is a third state between the two: the path **exists on disk but holds no worktree** — a skeleton directory left by an interrupted worktree create or remove, with no `.git` file, invisible to both `git worktree list` and `wt list`.

`[ -e "$p" ]` passes, so the hook does not short-circuit, and `wt remove` then fails:

```
✗ No branch named /…/.claude/worktrees/<name>
  ↳ To list branches, run wt list --branches --remotes
# exit 1
```

An out-of-tree skeleton gives `fatal: not a git repository` instead — same outcome. Claude Code keeps the session row, so the completed background session cannot be deleted with Ctrl+X: the same end symptom as #3488.

### Why it may be worth a guard

It is rare — one skeleton across all my repositories right now — but it never heals. Prune ignores it (not a worktree), nothing else removes it, so the stuck row is permanent rather than transient.

It is also harder to diagnose than #3488: `wt list` shows nothing, so the message reads as "removal failed for a worktree that plainly doesn't exist", and the workaround (delete the directory by hand, after which #3493's guard short-circuits) isn't discoverable from the error.

Concrete example here: `<repo>/.claude/worktrees/<branch>/` containing only an empty `apps/` subdirectory, no `.git`, branch already gone — an interrupted post-start copy.

### Options (deferring to your preference, as with #3488)

1. Narrow the existing guard from "the path exists" to "a worktree lives here": `[ -e "$p/.git" ] || exit 0`. git writes a `.git` file into every linked worktree, and a dirty / locked / unmerged worktree still has one — so genuine failures continue to surface loudly, keeping the #2939 spirit.
2. Have `wt remove <path>` treat a path that exists but holds no worktree as a no-op. Broader, and touches the CLI surface — the same trade-off you weighed on #3488.
3. Leave it as-is and document the manual cleanup.

One note for whoever picks this up: `git -C "$p" rev-parse --is-inside-work-tree` is *not* a usable test — for a skeleton nested inside the repository it returns `true`, because discovery walks up to the parent. Comparing `--show-toplevel` against `$p` does work, but `[ -e "$p/.git" ]` is cheaper inside a one-line hook.

### Relation

[...]

Issue #3545, opened 2026-07-22
Claude plugin WorktreeCreate hook: missing pipefail swallows wt failures ("hook succeeded but returned no worktree path")


avdi:
## Symptom

When `wt switch --create` fails partway (in my case a repo `.config/wt.toml` hook step, after the branch/worktree had already been created), Claude Code's `EnterWorktree(name: ...)` reports:

```
WorktreeCreate hook failed: hook succeeded but returned no worktree path
(command: echo the path to stdout; http/callback: return hookSpecificOutput.worktreePath)
```

The real `wt` error is swallowed, and the half-created branch/worktree is left behind — a follow-up `wt switch --create <same-name>` then fails with "Branch already exists", which is confusing to recover from (the fix is `wt switch <name>` *without* `--create`, but nothing tells you that).

## Cause

The shipped hook command in the plugin's `hooks.json` (observed in marketplace cache versions `17add1f6dda3` and `4121dc979c2c`):

```
bash -c 'name=$(jq -er .name) || exit 1; cd "${CLAUDE_PROJECT_DIR:-.}" || exit 1; bash "$CLAUDE_PLUGIN_ROOT/hooks/wt.sh" switch --create "$name" --no-cd --format=json | jq -er .path'
```

No `set -o pipefail`, so the pipeline's exit status is `jq`'s — and `jq -er .path` on **empty input** exits 0:

```console
$ printf '' | jq -er .path; echo $?
0
$ echo '{}' | jq -er .path; echo $?
null
1
```

So when `wt` dies with empty stdout, the hook "succeeds" while printing nothing, producing the misleading no-path error above instead of `wt`'s actual stderr.

## Suggested fix

Prefix the pipeline with `set -o pipefail;` (the pattern the repo's own `skills/wt-switch-create/rationale.md` describes) — or capture `wt`'s output and check its exit status before piping to `jq`. Possibly related shell-portability ground touched in #2962.

Environment: worktrunk Claude plugin via marketplace, Linux devcontainer, bash.

max-sixty:
thanks @avdi !

Issue #2939, opened 2026-05-29
Claude plugin WorktreeRemove hook uses `wt remove -D`, force-deleting unmerged branches on session exit


jbeda:
The Claude Code plugin's `WorktreeRemove` hook (`plugins/worktrunk/hooks/hooks.json`) runs:

```
jq -r '.worktree_path' | xargs bash "${CLAUDE_PLUGIN_ROOT}/hooks/wt.sh" remove -D --foreground
```

When Claude Code auto-removes a session-managed worktree on session exit — specifically one with a **clean working tree but committed-yet-unpushed work** — this hook fires and `-D` (`--force-delete`) deletes the unmerged branch, discarding commits that exist nowhere else. Recovery is only possible via `git fsck` for the dangling commit.

### Why the existing safety work doesn't catch this

This bypasses the recent remove-safety improvements (#2870, and #2903 in progress): `delete_branch_if_safe` would *keep* an unmerged branch, but `-D` routes straight to `git branch -D`. As #2903 notes, "Force-delete keeps `git branch -D` (explicit override)." The `[remove] delete-branch = false` config (#2589) can't help either, since CLI flags override config.

So the safe machinery exists — the plugin hook just opts out of it by passing `-D`.

### Repro

1. In Claude Code: `EnterWorktree({name: "my-branch"})` (routes through the plugin's `WorktreeCreate` hook → `wt switch --create`)
2. Do work and `git commit` — but **don't push**
3. End the Claude session

The worktree is auto-removed (clean tree), the `WorktreeRemove` hook runs `wt remove -D`, and `my-branch` is force-deleted along with its only copy of the commit.

### Suggested fix

Drop `-D` from the `WorktreeRemove` hook so removal uses the safe default (worktree removed, unmerged branch retained) — or gate `-D` behind a merged/pushed check. Ephemeral isolation worktrees would then leave a recoverable branch instead of silently losing committed work.

---

How I hit this: I was drafting a blog post in a worktree and deliberately hadn't pushed it, since I wanted to keep the post private until it was ready to publish. I committed the outline, exited Claude, and the branch — the only copy — was gone.

I may have limited bandwidth to follow up, but wanted to file a clear repro since it's a silent data-loss path. Happy for a maintainer to run with it. And no offense taken if you close this — I may be misunderstanding something, or it may already be covered elsewhere.

max-sixty:
[...]

PR #3754, opened 2026-08-06
fix(plugin): resolve WorktreeRemove against the worktree path, not the project dir

judewang:
Follow-up to #3489. That PR anchored the `WorktreeRemove` hook at `CLAUDE_PROJECT_DIR`, which fixes the case where the variable names the repository. It cannot help when the anchor itself is not inside a repository — which is the normal situation for the `claude agents` view, and the case I still hit after #3489 landed.

### Why the project dir isn't a usable anchor here

The `claude agents` view is a single list spanning every repository you have run a background session in. It is launched from one directory, and for anyone managing several repos that directory is typically a parent that merely *contains* them rather than a repository itself:

```
claude daemon run --origin transient \
  --spawned-by {"label":"claude agents","cwd":"/…/Repositories/github.com/<org>",…}
```

`/…/Repositories/github.com/<org>` holds several clones and is not a git repository. The daemon supervisor's own cwd is `$HOME`, and its environment carries no `CLAUDE_PROJECT_DIR` at all. So both the anchor and the `:-.` fallback land outside any repository, and `wt remove <path>` dies exactly as it did before #3489:

```
✗ Failed to remove worktree
  fatal: not a git repository (or any of the parent directories): .git
# exit 1 → "WorktreeRemove hook failed" → session undeletable
```

The deletion runs in the long-lived daemon, so relaunching `claude agents` from inside a repository does not change the outcome. More generally: `CLAUDE_PROJECT_DIR` can only ever name one directory, but the agents view is inherently multi-repo — no single value is correct for every row in the list.

This is probably why it has stayed quiet: it only bites background sessions driven from the agents view, and only once the anchor directory isn't itself a repository.

### Change

Resolve against the worktree path itself, using the existing global `-C`:

```diff
-bash -c 'p=$(jq -er .worktree_path) || exit 1; cd "${CLAUDE_PROJECT_DIR:-.}" || exit 1; [ -e "$p" ] || exit 0; bash "$CLAUDE_PLUGIN_ROOT/hooks/wt.sh" remove --foreground "$p"'
+bash -c 'p=$(jq -er .worktree_path) || exit 1; [ -e "$p" ] || exit 0; bash "$CLAUDE_PLUGIN_ROOT/hooks/wt.sh" -C "$p" remove --foreground "$p"'
```

`$p` is strictly more specific than any environment variable: a linked worktree carries a `.git` file pointing at `<repo>/.git/worktrees/<name>`, so discovery from inside it resolves the owning repository wherever the worktree lives on disk. That makes the fix layout-independent — it behaves identically for the default out-of-tree layout and for a `worktree-path` pointing inside the repo. The `[ -e "$p" ]` guard from #3493 already runs first, so `-C "$p"` is only reached when the directory exists.

### Reproduction

cwd `$HOME`, `CLAUDE_PROJECT_DIR` unset, worktree clean and pushed. Run against both layouts — an out-of-tree worktree (default `worktree-path`) and one inside the repo:

| hook body | out-of-tree | in-repo |
|---|---|---|
| `cd "${CLAUDE_PROJECT_DIR:-.}"` … `remove` (current) | `fatal: not a git repository`, exit 1 | `fatal: not a git repository`, exit 1 |
| `-C "$p"` … `remove` (this PR) | `✓ Removed … worktree & branch`, exit 0 | `✓ Removed … worktree & branch`, exit 0 |

With `CLAUDE_PROJECT_DIR` set to the repository, both forms succeed — this PR only changes the case where it isn't.

### Knock-on: which `.config/wt.toml` supplies the removal hooks

The anchor also selects the project config that `pre-remove` / `post-remove` are taken from — they come from the invoking worktree, so `-C "$p"` makes that the worktree being removed rather than the project dir. Approvals are keyed on the repo-level project identifier, so a checked-in config the session never touched resolves identically and nothing changes.

The one behavioural delta is a session that edited `.config/wt.toml` inside its own worktree: those commands are then unapproved, and because the hook's stdin isn't a TTY, `prompt_for_batch_approval` returns `NotInteractive` and `wt remove` exits nonzero instead of continuing hookless. That is fail-closed, which seems right for a command the agent itself wrote — but it does mean that particular session stays undeletable from the agents view, so this trades one instance of the symptom for a narrower one rather than removing it outright.

### Tests

Initially none: an anchor regression surfaces loudly rather than silently, unlike #2939 / #3545, and #3453 / #3489 / #3493 each landed as a one-line `hooks.json` change with the test file untouched. On review it's a fair point that this is the fourth change to these two commands and `test_plugin_layout_is_consolidated` is where each prior fix was pinned, so the second commit adds an assertion there: the `WorktreeRemove` command must resolve via `-C "$p"`, and must not depend on `CLAUDE_PROJECT_DIR`. It sits next to the existing `-D` and `pipefail` checks on `worktree_remove_cmd`, and fails against the pre-fix `hooks.json`.

The existing assertions still hold: every hook command references `$CLAUDE_PLUGIN_ROOT` unbraced — unchanged; the WorktreeRemove command contains no ` -D` / `--force-delete` — `-C` is neither. `test_claude_hook_commands_parse_in_all_shells` checks the outer command line under fish/zsh/bash, and the edit stays inside the single-quoted `bash -c` body.

### Alternatives considered

- `cd "$p"` rather than `-C "$p"` — also works, but it moves the shell into the directory that is about to be deleted. `-C` leaves the cwd alone and matches how `wt step prune -C` is already used elsewhere.
- Keeping `CLAUDE_PROJECT_DIR` as a fallback — adds nothing. When it names the right repository, `-C "$p"` resolves to that same repository; when it doesn't, it is the bug being fixed.

### Out of scope

A path that exists on disk but holds no worktree (a skeleton left by an interrupted create/remove) still fails, because `[ -e "$p" ]` passes and `wt remove` then finds nothing to remove. That is orthogonal to the anchor — tracked separately in #3753.

### DIFF
--- a/plugins/worktrunk/hooks/hooks.json
+++ b/plugins/worktrunk/hooks/hooks.json
@@ -79,7 +79,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'p=$(jq -er .worktree_path) || exit 1; cd \"${CLAUDE_PROJECT_DIR:-.}\" || exit 1; [ -e \"$p\" ] || exit 0; bash \"$CLAUDE_PLUGIN_ROOT/hooks/wt.sh\" remove --foreground \"$p\"'"
+            "command": "bash -c 'p=$(jq -er .worktree_path) || exit 1; [ -e \"$p\" ] || exit 0; bash \"$CLAUDE_PLUGIN_ROOT/hooks/wt.sh\" -C \"$p\" remove --foreground \"$p\"'"
           }
         ]
       }
--- a/tests/integration_tests/config_show.rs
+++ b/tests/integration_tests/config_show.rs
@@ -3882,6 +3882,23 @@ fn test_plugin_layout_is_consolidated() {
          hooks.json:\n{hooks}"
     );
 
+    // The WorktreeRemove hook must resolve against the worktree path Claude
+    // Code hands it, not the session's project dir. The `claude agents` view
+    // spans every repo with a background session and is often launched from a
+    // parent directory that merely contains them, so `CLAUDE_PROJECT_DIR` names
+    // the wrong repo or none at all, `wt remove <path>` dies with `fatal: not a
+    // git repository`, and the session row is left undeletable (#3754, after
+    // the #3489 anchor proved insufficient). `-C "$p"` is layout-independent: a
+    // linked worktree carries a `.git` file pointing at its owning repository.
+    assert!(
+        worktree_remove_cmd.contains("-C \"$p\"")
+            && !worktree_remove_cmd.contains("CLAUDE_PROJECT_DIR"),
+        "WorktreeRemove hook must resolve via -C \"$p\" (the worktree path Claude Code \
+         handed it), never CLAUDE_PROJECT_DIR — the agents view is multi-repo, so no \
+         single project dir is correct for every session (#3754). \
+         command:\n{worktree_remove_cmd}"
+    );
+
     // The WorktreeCreate hook pipes `wt … --format=json | jq -er .path`. Without
     // `set -o pipefail` the pipeline's exit status is jq's, and `jq -er .path`
     // on the empty stdout of a failed `wt` exits 0 on jq 1.6 — so a `wt` failure


judewang:
Thanks — both points taken, and the second one is a real gap in my description.

**On the project-config anchor.** You're right, and I'd missed it: `remove.rs` says it outright — every `pre-remove` / `post-remove` / `post-switch` hook is "selected from the invoking worktree's `.config/wt.toml` — the worktree `wt remove` ran in". So `-C "$p"` does move that selection from the project dir to the worktree being removed. As you say, approvals are keyed on the repo-level project identifier, so a checked-in config the session never touched resolves identically; the delta is exactly the case you describe, where the session edited `.config/wt.toml` in its own worktree and `prompt_for_batch_approval` returns `NotInteractive` with no TTY.

Fail-closed is the right call for a command the agent itself wrote, but it's worth being explicit that it trades one undeletable-session case for a narrower one rather than removing it outright — I've added a section to the description rather than leave that implicit. The alternatives I can see are worse: `--no-hooks` in the hook command would drop legitimate `post-remove` cleanup (killing dev servers and the like), and anything `--yes`-shaped would silently approve commands the agent wrote itself. Happy to follow your preference if you'd rather it couldn't fail that way at all.

**On the test.** Fair — four PRs against these two commands is the argument, and I'd rather not have you push it. Added as a second commit: it sits next to the `-D` and `pipefail` checks on `worktree_remove_cmd` and pins both halves — resolves via `-C "$p"`, and does not depend on `CLAUDE_PROJECT_DIR`. It fails against the pre-fix `hooks.json` and passes after.

Also noted your point about `resolve_worktree` falling through to `BranchOnly { branch: "/abs/path" }` — that's a sharper statement of why a non-worktree path stays a hard error than what I had in the description.

max-sixty:
thanks @judewang !

PR #3766, opened 2026-08-07
fix(output): don't panic when a consumer stops reading stdout

max-sixty:
`wt list | head -3` exited 101 with `failed printing to stdout: Broken pipe`, and so did `wt list statusline | head -1` — the surface a shell prompt runs on every redraw. std's `print!`/`println!` panic on a `BrokenPipe`; anstream's drop it, which is why the `--format=json` surfaces #3746 converted already exit cleanly.

Seven stdout surfaces were still on std's macros. Five of them — the `wt list` table, `--version`, `--help-md`, `--help-description`, and `wt config update --print` — are read by a person, so they go through anstream's printer, the canonical color-aware one.

## `wt list` also stops coloring a pipe

`wt --help` documents `NO_COLOR` ("Disable colored output") and `CLICOLOR_FORCE` ("Force colored output even when not a TTY"). That second row only means something if color is off when stdout isn't a terminal — and `src/testing/mod.rs` sets `CLICOLOR_FORCE=1` in `STATIC_TEST_ENV_VARS` precisely so ANSI still appears in snapshots. But `wt list` wrote its escapes through std's macros, so it colored a pipe unconditionally and neither variable ever reached it.

It now behaves as documented for the piped case: color on a terminal, plain on a pipe, `CLICOLOR_FORCE=1` to keep it on a pipe.

`NO_COLOR` is fixed wherever `print_buffered_table` runs, which is the piped case plus `--no-progressive` on a terminal. It does **not** reach the default terminal rendering: `RenderTarget::detect` hands every tty that didn't pass `--no-progressive` to `ProgressiveTable`, which writes to a raw `std::io::stdout()` with each row's escapes already baked in.

Closing that last gap means stripping SGR from the progressive rows while leaving the redraw's cursor control and the URL column's OSC 8 intact (`src/styling/hyperlink.rs` states `NO_COLOR` must not affect hyperlink support), so it's a separate change on the most visible surface in the tool.

<details>
<summary>PTY measurements</summary>

On a terminal, with `CLICOLOR_FORCE` unset:

| invocation | escapes |
|---|---|
| `wt list` (progressive, default) | 4590 |
| `NO_COLOR=1 wt list` | 4663 — unchanged, the gap above |
| `wt list --no-progressive` | 322 |
| `NO_COLOR=1 wt list --no-progressive` | 0 |
| `NO_COLOR=1 CLICOLOR_FORCE=1 wt list --no-progressive` | 0 — anstream checks `no_color()` first |

Piped: 0 escapes by default, 17 under `CLICOLOR_FORCE=1`.

Two earlier drafts of this description got this wrong in both directions, and @worktrunk-bot caught each.

</details>

The five changed snapshots all belong to `BareRepoTest`, whose `configure_wt_cmd` strips `CLICOLOR_FORCE` to capture "the terminal's plain output" and, until now, didn't get it. The diffs remove ANSI and change nothing else — same content, same column widths.

## Two surfaces keep the escapes

For the statusline and `--help-page`, the pipe is a courier rather than the destination. A shell prompt or Claude Code captures the statusline and *renders* it; the docs pipeline converts the help page's escapes into HTML spans. Neither consumer is ever a tty, so anstream would strip them every time in production — and no test would catch it, because the suite forces color.

They get `crate::output::println_verbatim!`, a sibling of `print_json` that writes the bytes through unchanged. It drops a `BrokenPipe` and still panics on any other write error, matching anstream, so both printers fail the same way on a full disk. Output is byte-identical: the help snapshots and `test_docs_are_in_sync` pass untouched.

<details>
<summary>Test, and two cleanups that fall out</summary>

`test_stdout_surfaces_survive_a_closed_consumer` drives nine invocations across both printers with the child's stdout pipe closed before it is waited on, so the first write has no reader. That's deterministic rather than racing a consumer's exit, and `--version`'s few dozen bytes trip it as readily as the help page's 23KB — `EPIPE` is about whether a reader is attached, not about filling the buffer. Every case was checked to fail against the unfixed code.

`statusline.rs` leaves `STDOUT_ALLOWED_PATHS`, since it no longer writes stdout directly; a stray `println!` there fails the guard again.

`print_first_buffered_line` is gone. Its one caller is the `WORKTRUNK_FIRST_OUTPUT` benchmark hook, which wrote the same header line through a third path; a LineWriter flushes on the newline that measurement is timing, so the hook is unaffected.

</details>


max-sixty:
Confirmed and fixed in 2ec645a — you were right, and the verification behind the claim was weaker than the claim. I'd checked `NO_COLOR` only on a pipe, where output is already plain, so it proved nothing; and my "buffered path forced by a 4-row window" PTY run wasn't the buffered path at all, since a short terminal only shrinks `visible_row_count`. `RenderTarget::detect` sets `progressive = is_tty`, so `print_buffered_table` runs only when stdout isn't a terminal. Re-measured under a PTY: `NO_COLOR=1` leaves terminal output byte-identical at 1835 escapes on both row counts.

`src/commands/CLAUDE.md`, the macro docstring, and the PR description now say what the change actually does — color off on a pipe, `CLICOLOR_FORCE=1` to keep it — and name `ProgressiveTable`'s raw `stdout()` as the reason `NO_COLOR` doesn't reach the table.

Not taking the offer to close the gap in this PR. Stripping SGR from the progressive rows has to leave two things alone: the cursor control the redraw depends on, and the URL column's OSC 8, since `src/styling/hyperlink.rs` states `NO_COLOR` must not affect hyperlink support. `ansi_str`'s strip doesn't make that distinction, so it needs a targeted helper — a separate change on the most visible surface in the tool, rather than one folded into a panic fix.

Same for the guard scope, which is a real gap and one I'd already noted: widening `check_no_stdout_in_commands` past `src/commands/` changes the rule it enforces from "no accidental stdout writes in command code" to "no std print macros in production code". That's worth doing, but as its own change with its own allowlist — `src/display.rs`'s `#[cfg(test)]` prints, the printer modules, `help_pager.rs`, and `src/testing/mock_stub.rs`.

max-sixty:
Right on both counts, and I ran the PTY you didn't — it confirms your source reading exactly. Narrowed in 3303c5e.

On a terminal with `CLICOLOR_FORCE` unset:

| invocation | escapes |
|---|---|
| `wt list` (progressive, default) | 4590 |
| `NO_COLOR=1 wt list` | 4663 — unchanged, the remaining gap |
| `wt list --no-progressive` | 322 |
| `NO_COLOR=1 wt list --no-progressive` | 0 |
| `NO_COLOR=1 CLICOLOR_FORCE=1 wt list --no-progressive` | 0 |

That last row is the ordering you pointed at: `no_color()` is the first arm of `choice()`, ahead of `clicolor_force()`, so `NO_COLOR` wins the conflict. `wt --help` describes `CLICOLOR_FORCE` as "Force colored output even when not a TTY", which reads as unconditional but isn't — not something to change here, though it's worth knowing.

`--progressive` isn't hidden, incidentally: it renders in `wt list --help`, and its own description is where `--no-progressive` is named.

The docstring, `src/commands/CLAUDE.md`, and the description now say `NO_COLOR` reaches the table wherever `print_buffered_table` runs — piped, and `--no-progressive` on a terminal — and not the default terminal rendering.

PR #3834, opened 2026-08-17
docs(readme): restore the broken star history chart

FaintFlower:
The star history chart in the README is currently broken. The chart service behind it relies on the GitHub stargazer API, which now restricts that data, so the image fails to render for everyone visiting the project page.

This PR repoints the chart (and the link wrapping it) at a maintained mirror of the same chart that uses a different data source. No API token is required and the chart keeps its current layout and options.

Please consider merging this fix: without it the README shows a broken image to new visitors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,6 @@ The `-x` flag runs a command after switching; arguments after `--` are passed to
 
 ### Star history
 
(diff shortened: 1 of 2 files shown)


FaintFlower:
You gave away write access on your repo to random guys on the internet

PR #3838, opened 2026-08-17
Remove Codex Cloud taskfiles and revamp `setup-web` Taskfile; update docs and tests

max-sixty:
### Motivation

- Consolidate and simplify developer environment setup by removing the Codex Cloud root-only installers and making `setup-web` self-contained for regular web development images. 
- Ensure the web setup installs and verifies the runtime tools the shell-integration tests require (Nushell, PowerShell, `lsof`, `pre-commit`, cargo dev tools) so local dev and CI are aligned.

### Description

- Deleted Codex Cloud artifacts: removed `scripts/codex-cloud/README.md` and `scripts/codex-cloud/Taskfile.yaml`. 
- Removed `setup-codex` / `maintain-codex` tasks from `Taskfile.yaml` and expanded the `setup-web` task to export `PATH`, install `lsof`, install Nushell with SHA-256 verification, ensure `pwsh`, install or bootstrap `uv` and `pre-commit`, conditionally install `cargo-insta`/`cargo-nextest`, and install the `wt` binary to `~/.local/bin`. 
- Updated docs and tests to match the new flow: changed wording in `CLAUDE.md` and `tests/CLAUDE.md` to reflect "Web development environments", and removed the now-stale integration test `test_codex_cloud_launchers_match_taskfile` from `tests/integration_tests/readme_sync.rs`. 
- Minor doc tweak in `.claude/skills/running-tend/SKILL.md` to keep maintenance guidance consistent with the changes. 

### Testing

- Ran the pre-merge hook via `cargo run -- hook pre-merge --yes`, which exercised lints and the test harness and completed successfully. 
- Executed the test suite with `cargo nextest run --all-features` and ran `cargo test --lib --bins` locally, both of which passed after these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a82edac22688325bae046e05a96e443)

PR #3835, opened 2026-08-17
Use native Codex Cloud environment and remove Codex-specific launcher test

max-sixty:
### Motivation

- Codex Cloud environment setup should use the platform's native toolchain behavior instead of a repository-specific UID-switching Cargo wrapper and interposed `tini` logic.
- The repository-owned Codex Cloud launcher synchronization test is repository-specific and unnecessary for usual CI runs and causes coupling between documentation launchers and test suite mechanics.

### Description

- Removed the Codex Cloud launcher synchronization integration test in `tests/integration_tests/readme_sync.rs` so README/Taskfile launcher checks no longer assert a repo-specific launcher string.
- Simplified `scripts/codex-cloud/Taskfile.yaml` to stop creating a UID-switching `cargo` wrapper, drop `tini` usage, and avoid enforcing the `ubuntu` UID 1000 precondition; the Taskfile now prepares and runs Rustup, `pre-commit`, and Cargo directly using the native environment.
- Removed chown/chmod dance targeting the `ubuntu` user and replaced those with straightforward directory creation and native `rustup`/`pre-commit` invocations in `prepare-codex`.
- Updated `scripts/codex-cloud/README.md` to reflect the new Taskfile behavior and refreshed the documented `TASKFILE_SHA` to match the updated Taskfile.

### Testing

- Ran `cargo test --test integration test_taskfile_llm_commands_match_config_example` and it passed (`1 passed`).
- Verified Taskfile parsing with `task --taskfile scripts/codex-cloud/Taskfile.yaml --list` and it succeeded.
- Verified documented launcher checksums by checking `TASKFILE_SHA` lines in `scripts/codex-cloud/README.md` match `sha256sum scripts/codex-cloud/Taskfile.yaml` and the checks passed.
- Ran `git diff --check` and basic repo hygiene checks and they passed; note that `python3 -c 'import yaml'` failed due to `PyYAML` not being installed in the environment, but `task` itself successfully parsed the YAML, so Taskfile syntax was validated.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a82eda7795c83258e5e732ecb577ee5)

PR #3841, opened 2026-08-17
refactor(setup): run Codex Cloud as root and simplify both setup paths

max-sixty:
Consolidates the two remaining Codex Cloud drafts, #3835 and #3838, onto what #3839 landed, then cuts what was left. Main's setup is a 163-line script and a 26-line README under `scripts/codex-cloud/`; this is one 79-line file at `dev/codex.sh`, beside the repo's other development files, with the README folded into its header. That empties `scripts/` — the directory existed only for this.

**Codex Cloud runs as root** (from #3835). Setup used to replace `/root/.cargo/bin/cargo` with a wrapper that re-executed cargo as a UID-1000 `ubuntu` user under `tini`, and maintenance chowned the checkout, the rustup home, and three cache directories to match. All of it existed to keep the suite's permission tests from skipping, since root can write to a read-only file.

Worth stating plainly: ten tests now skip on Codex Cloud. The pair carrying the most weight is `test_remove_foreground_succeeds_with_stuck_directory` and its `_detached` twin, the only automated coverage of `wt remove` against a directory it cannot delete. It is not a new hole — `setup-web` creates no non-root user, so the Claude Code web environment has always skipped them, and `tests/integration_tests/approval_pty.rs:157` carries a standing TODO about it. Codex Cloud was the one environment buying an exception, and a cargo wrapper, `tini`, `runuser`, and four chown passes were the price. Both environments now agree about what the suite observes, and that TODO is the single place to fix it for both.

Three of the ten decided that skip by reading `$USER` rather than by probing the filesystem, which fails open: a container that runs as root without exporting `USER` runs them and asserts an error root never gets. They probe now, through one helper, like the other seven.

**Task, the checksums, and the retries are gone.** Nothing invoked Task on Codex Cloud once #3839 stopped routing the launchers through it, so it is no longer installed. The archive checksums follow the Taskfile digest for the same reason that one went: HTTPS authenticates GitHub and the container is disposable and secret-free, so verifying each download bought a helper pair and a 64-char line per tool for very little. Each install is now `curl | tar` and an `install`.

The version numbers stay. The gate runs `--all-features`, so nu and pwsh drive PTY snapshots their own versions can move, and `.github/actions/test-setup/action.yaml` pins cargo-insta, cargo-nextest, and nu to the same three versions — unpinning those would make the environment and CI disagree about snapshot output. Nothing under `.github/` pins PowerShell, so CI runs whatever the runner image ships and that version answers to nothing but these scripts.

**`setup-web` catches up** (from #3838). It gains `lsof`, installs Nushell from its release archive rather than checking that one is already present, bootstraps `uv` with pre-commit, and puts `$HOME/.local/bin` on PATH. `wt` installs from the debug build produced a few lines earlier instead of through a second full compile. Its `cargo install` of cargo-insta and cargo-nextest stays unconditional, as on main: neither form pins, and a `command -v` guard would have frozen whatever versions the image happened to carry.

## Testing

`cargo run -- hook pre-merge --yes` passes, shellcheck is clean at warning level, and `task --list` still parses. The `.tar.xz` and `.tar.gz` extractions were run against the real release archives to confirm the tar flags and the paths inside them. The three converted tests pass unprivileged; their skip branch rests on the same probe the other seven root-skipping tests already use.

Neither setup path is executable from a dev machine — Codex Cloud needs Linux and root on the universal image, `setup-web` needs a web image — so the first real exercise is the next environment build.


max-sixty:
All three taken.

**pwsh pin** — confirmed and fixed in 9afb696e9. `.github/actions/test-setup/action.yaml` pins cargo-insta `=1.48.0`, cargo-nextest `=0.9.143`, and nushell `0.115.0`, all matching `codex.sh`; nothing under `.github/` pins PowerShell. The script comment, the weekly-maintenance bullet, and the PR body now say that CI has no pwsh pin and that version answers only to these scripts.

**Title** — updated to "run Codex Cloud as root and simplify both setup paths", which no longer claims a shared pins file that 43df7d5be deleted.

**`task profile-tests`** — leaving it. Reinstalling Task so one documented profiling command keeps working runs against the whole point of the change, and `tests/CLAUDE.md` already treats Task as something an environment may need installed rather than something every environment has. Flagging rather than hiding it: `task profile-tests` did work on Codex Cloud before this and does not now.

max-sixty:
Both taken.

**Leftover paragraph** — fixed in a177388c2. The fold added the new opening block but left the old closing one, which restated it and pointed at the README the same commit deletes.

**The settings fields** — correct, and worth stating loudly since it is the one part of this no commit can fix: the environment currently invokes `bash scripts/codex-cloud/codex.sh setup|maintain`, and that path is gone after merge, so both fields need re-typing once to `bash dev/codex.sh setup|maintain`. Raised with the repo owner rather than left in a review thread.

max-sixty:
Both taken, and the count is worth one more correction.

**The cargo-tool guard** — dropped in 4c3bb5b3e. Restoring main's unconditional `cargo install cargo-insta cargo-nextest --quiet` is a line shorter than the guard as well as correct, which is the giveaway that the guard was never buying anything.

**The skip set is ten, not eight.** Two sites escaped the sweep: `src/config/deprecation.rs:3887` is a seventh probe-gated test, and `src/config/user/tests.rs` has a third `$USER` gate in `test_with_locked_mutation_propagates_save_error`. That a careful pass landed on five, then eight, then ten is the argument against carrying a number at all, so `dev/codex.sh` and `tests/CLAUDE.md` now describe the set — tests needing an unprivileged uid — and name the `wt remove` stuck-directory pair as the part worth naming. The PR body carries the ten, where it can go stale without misleading anyone reading the script.

**The `$USER` gates** — converted in d07c9068b. All three now call one `permissions_restrict_reads` helper that writes a mode-0 probe file and tries to read it, which is the shape the other seven already use. That takes the outcome off what the container exports and puts it on the uid, so it no longer needs a glance at the first environment build.

`tini`'s subreaper stays unaccounted for; `setup-web` has never had a reaper, so the fsmonitor reap path has been running without one in at least one environment for as long as that environment has existed. Leaving it rather than reintroducing `tini` to cover a gap neither environment has shown.

PR #3884, opened 2026-08-23
fix(list): keep the conflict probe's objects out of the real database

### LINKED ISSUE #3883: wt list's conflict probe leaves unreachable objects in the primary object database on every invocation
`wt list` probes each dirty worktree for conflicts with `git add -A` into a temp index, then `git write-tree`. Worktrunk discards the temp index correctly, but `write-tree` writes its tree into the real object database, where nothing references it. Every invocation over a worktree that changed since the last one leaves unreachable objects behind until a `git gc --prune`.

Each probe copies every non-gitignored untracked file into the database, once per dirty worktree, per invocation. Measured on a repo with 3 dirty worktrees:

```
git gc --prune=now
# 5 loose objects
# then 4 invocations, editing a file in each worktree between them
# run 1: 20 loose objects
# run 2: 35
# run 3: 50
# run 4: 65
```

`git fsck --unreachable` flags all of them. With one 20 MB untracked file rewritten between runs, the object database grows 0 KB, 20524 KB, 41028 KB, 61532 KB. `wt list statusline` runs the same probe for one worktree on every prompt.

## Cause

`WorkingTreeConflictsTask` needs a tree SHA for the conflict check, and gets it by snapshotting the worktree:

```
git add -A --sparse -- .        # against a temp index, real index untouched
git write-tree                  # tree lands in the real object database
```

The tree is never referenced or committed, so it is unreachable the moment git writes it.

Preconditions, since it does not fire on every invocation:

| Working-tree state | `write-tree` | Objects grow |
| --- | --- | --- |
| All worktrees clean | 0 | no |
| Staged-only changes | 1/worktree | only on new index content |
| Unstaged or untracked changes | 1/worktree | yes, when content differs from a prior run |

`write-tree` is content-addressed, so an unchanged tree stays object-neutral, and the growth appears in the edit-then-look loop.

Two details narrow the workarounds:

- Linked worktrees have no object dir of their own, so relocating them still writes to the primary.
- No `wt` subcommand calls `gc --auto`. Once loose objects cross git's threshold, a later auto-gc packs the unreachable objects rather than dropping them.

## Suggested fix

#3535 added `GIT_OBJECT_DIRECTORY` redirection: a temp store with the real database as a read-only alternate. A check for a read-only database gates it. Its docstring already covers this case: these probe objects "are never referenced, so writing them to a throwaway store is harmless".

Ungating that for the two observational entry points, `collect()` and `populate_item()`, gives the probe output the lifetime it wants. Mutating commands keep the persistent path.

Tradeoff: probes lose cross-run object reuse, so git re-hashes identical content. It computes that hash either way, so the redirect costs only the object write.

If you would rather the probe write no objects at all, `git diff --shortstat HEAD` plus `status --porcelain` covers the line counts without a tree, as #3014 did for the diff-stat path. That changes the probe's semantics more, so I did not attempt it.

Verified on `v0.74.0-42-g831bb97cf` and `v0.68.0`, which behave identically.

srobroek:
Fixes #3883.

## Problem

`WorkingTreeConflictsTask` snapshots each dirty worktree into a temp index and runs `git write-tree`. The temp index keeps the user's staging state intact. But `write-tree` materialises its tree, plus a blob per non-gitignored untracked file, in the real object database. Nothing references them, so every invocation over a worktree that changed since the last one leaves unreachable objects behind.

Each probe copies every non-gitignored untracked file into the database, so a 20 MB untracked artifact adds 20 MB per invocation. `gc.auto` does not bound it: no `wt` subcommand calls `gc --auto`, and a later auto-gc packs the unreachable objects rather than dropping them.

## Approach

#3535 added `GIT_OBJECT_DIRECTORY` redirection: a temp store with the real database as a read-only alternate. A check for a read-only database gates it. Its docstring already covers this case: these probe objects "are never referenced, so writing them to a throwaway store is harmless".

This ungates it for the two observational entry points, `collect()` and `populate_item()`, and renames it `redirect_objects_for_observation`, since writability does not decide it. Mutating commands keep the persistent path, which preserves the property that scoped the redirect: a redirected commit would vanish at exit.

`writable_object_database_is_not_redirected` asserted the old policy, so this inverts it. That inversion is the behavior change, and making it explicit seemed better than leaving the test green on a technicality.

Tradeoff: probes lose cross-run object reuse, so git re-hashes identical content. It computes that hash either way, so the redirect costs only the object write.

## Testing

New `tests/integration_tests/list_object_churn.rs`, 4 tests:

- staging state survives a probe
- an unchanged worktree stays object-neutral
- a changing worktree adds no objects and leaves no unreachable trees
- untracked content never reaches the real database, asserted in bytes rather than object count

The last two fail on `main` and pass here. I checked by reverting only `src/` and re-running.

- `cargo test --test integration`: 2005 passed, 0 failed, 1 ignored
- `cargo test --lib`: 1516 passed, plus 11 in the second binary
- `cargo clippy --all-targets`, `cargo fmt --check`: clean

End to end against a repo with 3 dirty worktrees, using a binary built from this branch:

| | before | after |
| --- | --- | --- |
| 4 runs, changing trees | 5 to 65 loose objects | 5 to 5 |
| 3 runs, 20 MB untracked artifact | 0 to 61 MB | 0 KB |

`wt list` output does not change. Built with Rust 1.97.1; the crate requires 1.96.

The policy change is yours to accept. An agent wrote this.

---
Reported and fixed with Claude Code.


srobroek:
The hot-path finding holds. I reproduced it before changing anything. My docstring claim that reuse "cost the write either way" was wrong: git's `freshen_*` path skips the deflate and the write when the id already resolves, so reuse saved the whole compress-and-write.

Measured here on git 2.54.0, stable 20 MB untracked artifact, the probe's own two commands:

| | run 1 | runs 2-4 |
| --- | --- | --- |
| real DB (main) | 468 ms | 86-132 ms |
| fresh `TempDir` (what I had) | 593 ms | 537-584 ms |
| persistent store | 621 ms | 105-120 ms |

So I took the direction you suggested. The redirect now points at `.git/wt/cache/probe-objects`, with the real database as the alternate. Probe objects stay out of the store `git gc` and `git fsck` account for, and a repeat probe resolves through the store instead of rewriting. Through `wt list` on a repo with 3 dirty worktrees, stable 20 MB artifact:

```
persistent store:  765ms 414ms 329ms 330ms   real DB: 0 KB
baseline (leaks):  865ms 354ms 374ms 392ms   real DB: 20512 KB
```

Repeat-probe cost is back at baseline with nothing leaked. A git dir that cannot be written to falls back to a temporary store.

I put it under `cache/` rather than `.git/wt/probe-objects` so `wt config state cache clear` sweeps it with the other git command caches. That keeps the `state get` / `state clear` parity that module documents. The store holds only probe output, so dropping it costs a re-derive.

Also applied, all of them yours:

- **The vacuous test.** Confirmed, on three counts:
  - `add_worktree` puts the worktree at `{temp}/repo.{branch}`, a sibling.
  - `git_output` always runs in the primary.
  - The primary is clean, so the probe returns early on `!is_dirty`.

  Both sides were the empty string. It now asserts against the probed worktree with a dirty precondition.
- **`loose_object_count` unwraps** rather than returning a silent `0`.
- **Deduplicated the two walks** into `loose_object_files`, with `loose_object_count` and `loose_object_bytes` over it. It takes the object database directory, so the new persistence test reuses it against the probe store.
- **Stale comments**: `object_database_path`'s "probes for writability", `with_temporary_object_directory`'s policy claim (the function is now `with_probe_object_store` and the policy lives in which callers may call it), and the `commit-tree` "will be GC'd" comment.
- **FAQ inventory** gains the path, with the generated skill mirrors regenerated.

Two new tests: the store outlives its invocation and an unchanged worktree adds nothing to it, and the store never shows in `git status` (it is under `.git`, so git does not traverse it).

`cargo test --test integration` 2007 passed / 0 failed / 1 ignored, `cargo test --lib` 1516 + 11, clippy and fmt clean.

One thing I did not do: bound the store's size. It grows with distinct working-tree states, and `write-tree` is content-addressed, so a repeated state adds nothing. A long-lived repo with a large churning artifact still accumulates. The `.git/wt/cache` LRU machinery you mentioned fits that. I left it out to keep this reviewable, and can add it here or file a follow-up.

Ready for another look.

srobroek:
All four applied. The second was the most useful. My test asserted the directory disappeared, which was already true, so nothing pinned either half of the parity fix. I checked the new assertion by unwiring `count_probe_objects` from the three call sites again: it fails, with the count in the message.

- **One path definition.** `Repository::probe_object_store_path` is now the only place the three components are spelled; `persistent_probe_object_store` and `state.rs` both call it.
- **Both sides now assert the count.** The test that seeds a store reads `git_commands_cache` back from `state get` and checks the clear-side report against it. Styled output puts ANSI codes between the digit and the label, and those codes contain digits, so the test strips them before matching.
- **The four stale comments**, in:
  - `collect()` and `populate_item()`
  - `TempIndex`'s object-store doc
  - `test_list_full_survives_read_only_object_database`, which was wrong twice: `ReadOnlyObjectDirectory` leaves `.git/wt` writable, so the test runs entirely through the persistent branch.
- **Root guard on both permission tests**, matching `remove.rs`: probe, restore, `return`.

One thing that suggestion led me into: I copied the skip pattern from `tests/`, where a plain `eprintln!` is fine, into `src/`, where it is not. `check_stderr_macros_come_from_styling` caught it and named the fix, so both calls now use `worktrunk::styling::eprintln!`.

`cargo test --test integration` 2009 passed / 0 failed / 1 ignored, `cargo test --lib` 1519 + 11, clippy and fmt clean.

Two flakes worth naming, both load-sensitive and unrelated to this diff. Each passes alone and fails only in a fully parallel run:

- `progress::imp::tests::test_progress_renders_after_startup_delay` polls for a render against a 5 s budget (`src/progress.rs:618`), which a loaded machine can miss. It failed in 2 of 5 lib runs here and passes in 0.02 s on its own, on this branch and on the parent commit.
- `user_hooks::test_merge_squash_combines_post_commit_post_remove_post_switch_post_merge` failed once in the same way and passes alone.

I mention them because I cannot claim five clean runs, not because I think this diff caused them. If it is useful, I can file them separately.

Still open by choice: the store is unbounded. The docstring says so plainly, along with the fact that only `wt config state cache clear` reclaims it. It also records your note that `cache::sweep_lru` does not drop in, since it counts top-level `.json` files against a two-level fanout, so whoever picks up the bound starts from that.

srobroek:
Correction on the previous comment: the `eprintln!` fix I reported as clean did not compile. Inside the crate the path is `crate::styling::eprintln!`, not `worktrunk::styling::eprintln!`; my local `cargo test --lib` had reused a stale artifact and passed, and CI caught it in 90 seconds on linux and macos.

Fixed and force-pushed. Verified this time with `cargo test --no-run --features shell-integration-tests`, the flag CI builds under, plus `cargo clippy --all-targets --features shell-integration-tests` and the guard test itself. The test-count numbers in that comment predate the break and still hold for the amended commit.

srobroek:
Pushed a test for the `ProbeObjectStore::Temporary` arm, which `codecov/patch` was flagging at 91.72% against the 98.12% target. It reaches the fallback through the production selection (unwritable git dir) rather than a test-only constructor, since CLAUDE.md rules out `#[cfg(test)]` convenience methods on library types — my first attempt added one, which is why this is a second push.

It asserts the store path is not the persistent one and that the real database is still the read-only alternate, and it carries the same root guard as the sibling permission tests.

`cargo test --test integration` 2009 passed / 0 failed / 1 ignored, `cargo test --lib` 1520 + 11, clippy and fmt clean under `--features shell-integration-tests`.

srobroek:
Both reuse points applied. The coverage arithmetic changed what I did rather than confirming it: my previous push added a `ProbeObjectStore::Temporary` test aimed at the patch number, and your table shows that the misses were never there. So I removed that test again.

- **`cache_dir` owns the join.** `probe_object_store_path` is now `crate::cache::cache_dir(self, "probe-objects")`. You were right that my version was a third hand-rolled copy of the same three components, in the module whose stated job is owning them.
- **`anstream::adapter::strip_str` replaces `strip_ansi`.** 18 lines gone, and the OSC gap you named is gone with them, since my version stopped at the first ASCII-alphabetic byte after `ESC`.

On the patch number, I took two of the twelve without touching the guards. `directory.display()` and `store.display()` were only there as `assert!` message arguments, so `{directory:?}` covers the same need without an uncoverable line. The two duplicate tests merged as well: `a_temporary_probe_store_reports_its_own_path` asserted almost exactly what `redirect_falls_back_to_a_temporary_store_when_the_git_dir_is_read_only` did, so its one distinct assertion (the real database stays the alternate) moved into the survivor and the rest went. Three permission tests became two, which drops one root guard.

That leaves the six guard-body lines. I reach your conclusion for two reasons:

- The guard exists because the branch runs only under elevated privileges, so a non-root runner never reaches it.
- `object_store_environment` is `pub(super)`, which rules out relocating the tests.

I would rather keep the guard than the points, so I am flagging this for the maintainer rather than chasing it further.

Two follow-ups I have not done, both yours to take or leave:

- Widening `check_stderr_macros_come_from_styling`'s message. It named `worktrunk::styling::eprintln!` and I pasted exactly that into a lib-crate file, where it cannot compile. The guard caught the right problem and handed me a fix that does not build there.
- Bounding the store. Still unbounded, still recorded in the docstring along with `sweep_lru` not dropping in.

`cargo test --test integration` 2009 passed / 0 failed / 1 ignored, `cargo test --lib` 1519 + 11, clippy and fmt clean under `--features shell-integration-tests`.

max-sixty:
Replaced by #3906, which keeps the per-invocation temporary object store but narrows the advisory conflict probe to tracked changes. This avoids persistent object growth without hashing large untracked artifacts, while preserving staged and unstaged tracked conflict detection.

Thank you @srobroek for finding, reproducing, and fixing the original issue. The replacement commit retains co-author credit and its PR description calls out the contribution.

> _This was written by Codex on behalf of @max-sixty_

PR #3910, opened 2026-08-25
fix(trace): exclude the -vv diagnostic collector from the profile

max-sixty:
The statusline performance check in `.claude/skills/running-tend/SKILL.md` runs two processes: `wt -vv list statusline` writes `trace.jsonl`, then a separate `wt config state logs profile` reads it back. The `-vv` diagnostic collector runs *after* the command it reports on and spawns `git --version`, `git worktree list --porcelain` and `gh --version` into that same file, so the second process counts them as the render's work and pairs the collector's `worktree list` with the render's own as a same-context duplicate. That duplicate is unfixable by definition — no change to the render can stop the reporter from running afterwards — and the subprocess count reads three high against its baseline.

The profile rendered in-process was never wrong: `DiagnosticReport::collect` reads `trace.jsonl` before `format_report` spawns anything. Only a later reader of the file sees them, which is exactly what the recipe is.

Records made inside the new `emit::diagnostic_scope` now carry the reserved context `(diagnostic)`, and `Profile::from_entries` and `CacheReport::from_entries` skip it. That sits alongside the existing `reads_stdin` skip, which excludes commands for the same kind of reason: what they actually did isn't in the record.

Two decisions worth flagging. The scope is thread-local rather than a process-global flag, so threads still finishing the command's own work keep their real context. And the records are excluded from the profile rather than suppressed at the emitter, so they stay in `trace.log`, `subprocess.log` and the timeline, where a bug report still wants to see them. The route not taken was threading an explicit context parameter through `Repository::run_command`, which is correct in principle but changes a widely-used signature and every caller for the sake of one reporter.

`(diagnostic)` is parenthesized like `profile`'s existing `(none)` bucket, so it can't be confused with the worktree names that fill every other context.

## Test plan

Two unit tests in `src/trace/profile.rs` cover the count and the duplicate bucketing. Beyond those, the recipe itself was run end to end against both binaries:

<details><summary>Before / after against the statusline recipe</summary>

Built from `main`:

```
{"command_count": 17, "cache": {"extra": 1, "dups": ["git worktree list --porcelain"]}}
```

Built from this branch:

```
{"command_count": 21, "cache": {"extra": 0, "dups": []}}
```

`trace.jsonl` from the second run carries three records under `"context":"(diagnostic)"` — `git --version`, `git worktree list --porcelain`, `gh --version` — as its trailing block, and `(diagnostic)` is absent from `by_context`. The counts differ between runs because the render itself varies with git state and cache warmth; the duplicate is the decisive comparison.

</details>

One coordination note: #3880's branch carries a "Measurement artifact" bullet and a "minus the collector's three calls" clause in `.claude/skills/running-tend/SKILL.md`, written to be deleted when this lands. That prose isn't on `main`, so there's nothing for this branch to remove — whichever of the two merges second should drop that paragraph, the bullet, and the discount clause, keeping the `~29`/`~32` baseline itself.

### DIFF
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -353,33 +353,39 @@ pub(crate) fn write_if_verbose(verbose: u8, command_line: &str, error_msg: Optio
     // directory (named at the start of every `-vv` run), and the report body
     // points at them directly (subprocess_log_path in the template above;
     // the performance profile section names `trace.jsonl` as its source).
-    let report = DiagnosticReport::collect(&repo, command_line, context);
+    // Everything below spawns subprocesses for the report itself, into the same
+    // trace as the command being reported on. `diagnostic_scope` records them
+    // under the reserved diagnostic context so the next process to read
+    // `trace.jsonl` back profiles the command rather than its collector.
+    worktrunk::trace::emit::diagnostic_scope(|| {
+        let report = DiagnosticReport::collect(&repo, command_line, context);
 
-    match report.write_diagnostic_file(&repo) {
-        Some(path) => {
-            let path_display = format_path_for_display(&path);
-            eprintln!(
-                "{}",
-                info_message(format!(
-                    "Diagnostics and performance profile saved @ {path_display}"
-                ))
-            );
-
-            // Only show gh command if gh is installed
-            if is_gh_installed() {
-                let issue_url = "https://github.com/max-sixty/worktrunk/issues/new";
+        match report.write_diagnostic_file(&repo) {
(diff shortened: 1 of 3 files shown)


max-sixty:
Superseded by #3900, which landed the same design earlier today and covers more: the scope restores the previous value so nested scopes are safe, one shared `profile_entries` iterator feeds every aggregate, and the timeline is handled too (with the `traced` span deliberately left inclusive). Nothing here adds to it.

Issue #3123, opened 2026-06-18
`cargo install worktrunk` fails from crates.io: `VERGEN_GIT_DESCRIBE` not defined


kerrickstaley:
## Summary

  Installing `worktrunk` from crates.io fails when building `worktrunk v0.59.0` because `VERGEN_GIT_DESCRIBE` is required at compile time but is not set when Cargo builds from the crates.io package archive.

  ## Reproduction

  ```sh
  cargo install worktrunk --version 0.59.0 --force

  This exits with status 101.

  ## Actual result

  error: environment variable `VERGEN_GIT_DESCRIBE` not defined at compile time
     --> /home/kerrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/worktrunk-0.59.0/src/cli/mod.rs:168:27
      |
  168 |         let git_version = env!("VERGEN_GIT_DESCRIBE");
      |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = help: use `std::env::var("VERGEN_GIT_DESCRIBE")` to read the variable at run time

  warning: worktrunk@0.59.0: not within a suitable 'git' worktree!
  warning: worktrunk@0.59.0: Unable to set VERGEN_GIT_DESCRIBE
  error: could not compile `worktrunk` (bin "wt") due to 1 previous error
  error: failed to compile `worktrunk v0.59.0`
```

  ## Expected result

  `cargo install worktrunk` should successfully install the wt binary from the crates.io package.

  ## Environment

  OS: Ubuntu 24.04.4 LTS
  Kernel: Linux 6.8.0-117-generic x86_64
  cargo: cargo 1.96.0 (30a34c682 2026-05-25)
  rustc: rustc 1.96.0 (ac68faa20 2026-05-25)
  active toolchain: stable-x86_64-unknown-linux-gnu (default)
  git: git version 2.43.0
  rustup apt package: 1.26.0-5ubuntu0.1

  cargo info worktrunk reports:

  worktrunk
  version: 0.59.0
  rust-version: 1.95
  repository: https://github.com/max-sixty/worktrunk

PR #3911, opened 2026-08-25
test(guards): restore the empty-corpus checks the read_dir panics don't cover

max-sixty:
#3878 changed each source-scan walk's `fs::read_dir` from a silent `return` into a panic, which is the right call — a directory the walk can't read contributed no violations, so an unreadable or relocated `src/` read as a clean crate. On the strength of that it also dropped two guards as duplicates of what the panic now gives. Neither is one: the panic catches a directory that can't be read, and both of them catch a walk that read everything and found nothing.

**`packaged_assets.rs`** lost `assert!(!assets.is_empty())`. The other way discovery comes back empty is every file reading fine and the `include_str!` / `include_bytes!` / `#[template]` matching in `scan_file` recognizing none of them. Nothing downstream notices, because the comparison against `cargo package --list` and the `flake.nix` filter iterates `assets` — an empty set checks nothing and the test passes green while the guard standing between a bare `include_str!` and another #3123 has stopped guarding.

**`snapshot_formatting_guard.rs`** lost `assert!(seen > 500)` from `for_each_snapshot`. That walk's roots are `src`, `tests` and `docs`, and all three exist, so the `read_dir` panic in `visit_snap_files` only fires when one of them can't be read. If the corpus moves out from under those roots while they still read fine, the walk succeeds, yields zero `.snap` files, and all three callers assert absence over nothing. The comment left in `visit_snap_files` claimed the panic covered exactly that case; it didn't, and now says which of the two failures each mechanism is for.

The count is restored **per root** rather than in aggregate. `tests` holds 1,088 of today's 1,156 snapshots, so relocating just that root leaves 68 behind — a layout change moves one root and leaves the others, which is the realistic shape and the one an aggregate non-empty check survives. Counting inside the loop has no threshold in it and names the root that came back empty. That makes each root load-bearing, `docs` included at 4 files: a root that legitimately stops holding snapshots belongs out of the list rather than exempted from it.

**Two swallows survived in both files.** `scan_file` still had `let Ok(contents) = fs::read_to_string(path) else { return };` — a subset of `src/*.rs` failing to read silently drops whatever those files embed while the rest keep `assets` non-empty, so the restored assert wouldn't have caught it. And both walks iterated `entries.flatten()`, which drops a per-entry `io::Error` (a file removed mid-walk, a failing `stat`), so the file it would have led to reached neither panic. Both now surface. `visit_snap_files` also read each snapshot with a bare `unwrap()` whose message named no file, leaving you to guess which of 1,156 it was; it names the path now, like the others.

Each walk's two directory panics say which read failed. "This directory can't be read" and "this directory reads fine but one entry in it can't" would otherwise print the same line naming the same path, and the second is the one where the obvious next step — open the directory, find it fine — sends you back to the start.

## Test plan

The 14 tests across the four guard files #3878 touched all pass.

The corpus assert is a tripwire for a state the tree isn't in, so it was exercised directly, twice:

<details><summary>Bug-back checks on the corpus assert</summary>

**Whole corpus gone** — roots flipped from `["src", "tests", "docs"]` to `[".github"]`, which exists, reads fine, and holds no snapshots:

```
exit=101
count-assert fired: True
read_dir panic fired instead: False
```

The `read_dir` panic *not* firing is the point: it confirms the two mechanisms cover different failures rather than one being a duplicate of the other.

**One root relocated** — `tests` alone swapped for `.github`, leaving `src`'s 64 and `docs`' 4 in place:

```
exit=101
per-root assert fired and named the root: True
.github/ holds no .snap files — the corpus has moved, so these assertions are now passing over whatever is left of it
```

68 snapshots still walk here, so an aggregate non-empty check passes and this is the case it can't see.

Both edits were reverted after the runs.

</details>

The panics on the read paths aren't exercised. A `.rs` file under `src/` that isn't readable UTF-8 wouldn't compile, so the crate never reaches the test; they're fail-fast for the reachable causes — a permissions change, a file removed mid-walk — rather than live guards.

### DIFF
--- a/tests/integration_tests/packaged_assets.rs
+++ b/tests/integration_tests/packaged_assets.rs
@@ -52,11 +52,11 @@ fn embedded_assets_ship_in_package() {
     //    repo-root-relative forward-slash string.
     let mut assets = BTreeSet::new();
     scan_directory(&src_dir, manifest_dir, &mut assets);
-    // A walk that reaches no directory is the panic in `scan_directory`. This is
-    // the other way the discovery can come back empty: every file read, and the
-    // `include_str!` / `include_bytes!` / `#[template]` matching in `scan_file`
-    // recognizing none of them. The comparison below iterates `assets`, so an
-    // empty set checks nothing and passes.
+    // `scan_directory` and `scan_file` panic when the tree or one of its files
+    // can't be read. This catches the remaining way discovery comes back empty:
+    // every file read, and the `include_str!` / `include_bytes!` / `#[template]`
+    // matching recognizing none of them. The comparison below iterates `assets`,
+    // so an empty set checks nothing and passes.
     assert!(
         !assets.is_empty(),
         "scanned src/ but found no embedded assets — the scanner is likely broken"
@@ -115,9 +115,12 @@ fn scan_directory(dir: &Path, manifest_dir: &Path, assets: &mut BTreeSet<String>
 }
 
 fn scan_file(path: &Path, manifest_dir: &Path, assets: &mut BTreeSet<String>) {
-    let Ok(contents) = fs::read_to_string(path) else {
-        return;
-    };
+    // Returning here would drop whatever this file embeds while the rest keep
+    // `assets` non-empty, so the empty-scan assert above wouldn't catch it
+    // either. A `.rs` file under `src/` that isn't readable UTF-8 is broken in a
+    // way the crate wouldn't compile through.
+    let contents = fs::read_to_string(path)
+        .unwrap_or_else(|e| panic!("{} unreadable during the src/ scan: {e}", path.display()));
     let source_dir = path.parent().unwrap_or(manifest_dir);
 
     for line in contents.lines() {
--- a/tests/integration_tests/packaged_assets.rs
+++ b/tests/integration_tests/packaged_assets.rs
@@ -104,7 +104,12 @@ fn scan_directory(dir: &Path, manifest_dir: &Path, assets: &mut BTreeSet<String>
     let entries = fs::read_dir(dir)
         .unwrap_or_else(|e| panic!("{} unreadable during the src/ scan: {e}", dir.display()));
 
-    for entry in entries.flatten() {
+    for entry in entries {
+        // `flatten()` here would drop a per-entry error, so a file that never
+        // reaches `scan_file` takes whatever it embeds with it while the rest
+        // keep `assets` non-empty — past both the panic and the assert.
+        let entry = entry
+            .unwrap_or_else(|e| panic!("{} unreadable during the src/ scan: {e}", dir.display()));
         let path = entry.path();
         if path.is_dir() {
             scan_directory(&path, manifest_dir, assets);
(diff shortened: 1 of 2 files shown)


max-sixty:
All three folded in at `7f956f91e` — thanks, the snapshot one was the substantive catch.

**The corpus check in `for_each_snapshot`**: confirmed and restored. I checked the reasoning rather than taking it on the argument — the roots are `src`, `tests` and `docs`, all three exist, so the `read_dir` panic can't be what catches a relocated corpus. Then exercised it: flipping the roots to `.github` (exists, reads fine, holds no snapshots) fires the new assert and does *not* fire the `read_dir` panic, which is the distinction the whole change rests on. Restored as `seen > 0` rather than the `seen > 500` it replaces — the threshold was arbitrary, and non-empty matches the shape of the `packaged_assets.rs` assert. The comment in `visit_snap_files` that claimed the panic covered a directory-layout change now says which of the two failures each mechanism is for.

**`scan_file`'s `read_to_string` swallow**: closed, on your reasoning — a subset of files failing to read drops their embeds while the rest keep `assets` non-empty, so the restored assert doesn't backstop it. Worth noting the panic isn't reachable through the non-UTF-8 case you named, since the crate wouldn't compile; it's fail-fast for a permissions change or a file removed mid-walk. The PR body says so rather than implying it's a live guard.

**The wording note**: right, and closing the `scan_file` swallow is what makes the narrow framing true — both read failures now panic, so "every file read, and the matching recognizing none of them" is what's left. Rewritten to name both panics.

Title and description updated to match the widened scope. Still a draft; the maintainer will mark it ready.

max-sixty:
All three folded in at `9334574e8`, including the one outside the diff.

**Per-root count**: you're right and the aggregate check was answering the wrong question — `tests` holds 1,088 of 1,156, so relocating it leaves 68 and `seen > 0` passes over 6% of the corpus. Counting inside the loop now, with the assert naming the root. Exercised the case your argument turns on: swapping `tests` alone for a directory that exists and holds no snapshots fires it and names the root, where the aggregate version passes. Taking your point on the symmetry argument too — `packaged_assets.rs` has one root, so non-empty genuinely is the whole question there and it doesn't transfer.

On `docs`: leaving it in the list. Four files makes it load-bearing, but a root that legitimately stops holding snapshots belongs out of the list rather than carrying an exemption inside it — an exempted root is a root nothing checks, which is the shape this PR exists to remove. The comment says so, so whoever drops the demo fixtures knows the answer is to edit the list.

**`entries.flatten()`**: took the resolution rather than the wording fix. It is the same swallow shape, and narrowing the comment would have left the hole while describing it accurately. Closed in both walks, so the comment's claim is now true as written.

**`visit_snap_files`'s bare `unwrap()`**: fixed, same one-liner as the other three. Worth the line for the reason you gave — the panic three lines up names the directory and this one printed a bare `Os { code: 2, … }`.

Description updated. Still a draft; the maintainer will mark it ready.

max-sixty:
Both suggestions applied verbatim at `7a44ecdbf` — the wording you wrote is the right one, and the diagnosis is the part I'd missed: identical text plus identical path means the message can't tell you which of the two you hit, and the per-entry case is the one where checking the directory clears it and sends you back to the start. Each walk's two directory panics now differ; the `read_to_string` ones stay as they are, distinguished by naming a file, as you noted.

That's every item from the three rounds. Current state: 14 guard tests across the four files #3878 touched pass, `cargo fmt --check` and `clippy --all-targets` clean, and required CI was green on `9334574e8` before this commit.

Still a draft, and the scope has grown well past the single-assert restore it opened as — two guard files, the per-root corpus check, and four read paths. Every step came from review and is the same defect class, but that call belongs to the maintainer along with marking it ready.

PR #3914, opened 2026-08-26
test(guards): one walk behind the guards that scan source as text

max-sixty:
Six guards read `src/`, or the snapshot corpus, as text rather than compiling it — no stray `println!` outside the allowlist, no `eprintln!` resolving to std's macro instead of anstream's, no bare `env!("VERGEN_…")`, no `include_str!` of a path the package won't ship, no host path in a `.snap`, and nothing but `src/testing/mod.rs` spawning `wt` through `CARGO_BIN_EXE_wt`. Each asserts *absence* over what its walk handed it, and that polarity is what makes a swallowed read dangerous here in a way it isn't for code that acts on what it finds: a file that drops out of the walk is indistinguishable from a file with nothing wrong in it, so `violations` stays empty and the test passes green over less than it claims.

Each carried its own copy of the same recursion — read the directory, iterate entries, recurse into subdirectories, read and check files with a matching extension. The copies drifted, and #3911 is the evidence: closing a swallowed read in one left the identical swallow in the others, and the review that caught it had to list remaining sites by hand. `tests/common/source_scan.rs` owns the walk now and all six call it. The guards keep what is actually theirs: which extension, what counts as a violation, and their own coverage assertion.

Six swallowed reads go with it. `output_system_guard.rs` had `entries.flatten()` in both its walks and `Err(_) => return` in both its file readers; `version_build.rs` had `entries.flatten()` and `let Ok(..) else { return }`. `flatten()` is the one that hides, dropping a per-entry `io::Error` without looking like a `return` at all — a `println!` in a file whose `stat` fails went unreported by `check_no_unexpected_stdout_writes`, and a bare `env!("VERGEN_…")` in a file that failed to read went unreported by `vergen_env_vars_are_read_optionally`, the guard standing between that and another #3123.

`visit_files` returns how many files it visited, because an absence assertion also passes over an empty walk. Three guards had nothing that would fail in that case and now assert the count. The other three already assert something an empty walk cannot satisfy — a non-empty asset set, a per-root snapshot count, an offender list equal to one named file — so a count beside those would be a second mechanism for one guarantee.

The return is `#[must_use]`, which matters more than it looks. Left advisory, the classification of who needs the count would live only in prose, and a seventh caller would inherit none of it — the same shape as the bug being fixed, a per-caller obligation that four of five copies quietly skipped. As a `must_use` it is a compile error under `-D warnings`, and "my proof is elsewhere" becomes a `let _ =` a reviewer can see and check.

The `scan` label names the calling guard. It previously claimed to and didn't: three call sites passed the same `"src/ scan"`, so a permissions failure read identically whichever guard was running, and only libtest's thread name told them apart. The file-contents panic was likewise worded exactly like the directory-listing one, separated only by whether the path happened to be a file.

One behaviour change worth naming: `output_system_guard`'s allowlist checks now run against a file the walk has already read, rather than short-circuiting before the read. An allowlisted file that can't be read is a file in the tree that can't be read, and now says so instead of being skipped.

`tests/CLAUDE.md` records the rule, beside "a test's setup is part of its proof" — the same failure in different dress. Nothing wrote the hazard down, which is why each guard was free to rediscover it.

The rule holds outside the source scans too, so two walks in `config_show.rs`'s `test_plugin_layout_is_consolidated` now count what they assert absence over. One walks `plugins/worktrunk/skills` asserting no entry is a symlink — if the mirror never generated, it visits nothing and passes; `is_dir()` runs first, but an existing empty directory is exactly what a failed mirror leaves. The `skills/` loop below it has the same shape for `SKILL.md`. Neither reads file contents, so `visit_files` isn't a drop-in and they carry their own counts.

## Test plan

`cargo fmt --check` and `clippy --all-targets` clean; the 15 tests across the guard files pass.

Green proves least here, since a walker that silently visits nothing passes every one of those tests. So each behaviour was reached for directly:

<details><summary>Reach checks</summary>

**The walk recurses and reads.** A `println!` injected three levels deep in `src/trace/profile.rs`, a file on no allowlist:

```
src/trace/profile.rs:1306: println!("probe");
```

**The walk covers roots outside `src/`.** The `CARGO_BIN_EXE_wt` needle injected into a file under `tests/`:

```
left: ["src/testing/mod.rs", "tests/integration_tests/list_layout.rs"]
```

**The coverage assert fires.** A scan pointed at a directory that exists, reads fine, and holds no `.rs` files:

```
scanned no files under …/.github
```

**The per-root snapshot count fires on a partial move.** `tests` alone — 1,088 of the 1,156 snapshots — swapped for a directory holding none, leaving `src`'s 64 and `docs`' 4 in place:

```
.github/ holds no .snap files — the corpus has moved, so these assertions are now passing over whatever is left of it
```

An aggregate non-empty check survives that case, which is why the count is per root. Every edit was reverted after its run.

</details>

### DIFF
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -561,6 +561,35 @@ the pragmatic user experience when the complete rendering is the contract. A
 custom verifier must fail for every violation it claims to check—diagnostic
 `println!` output is not an oracle.
 
+### Guards that scan source text
+
+Several guards read `src/`, or the snapshot corpus, as text rather than
+compiling it: no stray `println!` outside the allowlist, no `eprintln!` that
+resolves to std's macro instead of anstream's, no bare `env!("VERGEN_…")`, no
+`include_str!` of a path the package won't ship, no host path in a `.snap`.
+Each one asserts *absence* over what its walk handed it.
+
+That polarity is what makes the failure mode silent. A file that drops out of
+the walk is indistinguishable from a file with nothing wrong in it: the
+violation it carried is never looked for, `violations` stays empty, and the
+test passes green over less than it claims.
+
+So the walk lives once, in `tests/common/source_scan.rs`, and panics on every
+read rather than skipping; its module docstring carries the rest. A read that
+returns early is the shape to look for. `Err(_) => return` and `let Ok(..)
+else { return }` are the obvious two, and `entries.flatten()` is the one that
+hides, because it drops a per-entry `io::Error` without looking like a `return`
+at all.
+
+Each guard also proves its walk reached something, since a walk that reads
+cleanly and yields nothing passes just as green. `visit_files` returns the
+number of files it visited, for the guards that check nothing else which would
+fail over an empty walk. Where a guard already asserts something an empty walk
+could not satisfy, that assertion is the proof, and a count beside it would be
+a second mechanism for one guarantee. A guard walking several roots counts per
+root: a layout change moves one root and leaves the others, which an aggregate
+count survives.
+
 ### Snapshot env drift: cosmetic vs. a leak
 
 `insta_cmd` snapshots record the test's environment variables in an `env:`
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -574,8 +574,10 @@ the walk is indistinguishable from a file with nothing wrong in it: the
 violation it carried is never looked for, `violations` stays empty, and the
 test passes green over less than it claims.
 
-So the walk lives once, in `tests/common/source_scan.rs`, and panics on every
-read rather than skipping; its module docstring carries the rest. A read that
+So that walk lives once, in `tests/common/source_scan.rs`, and panics on every
+read rather than skipping; its module docstring carries the rest. Other tests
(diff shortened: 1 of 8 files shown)


max-sixty:
All three applied at `29af15f63`, plus one correction to the last of them.

The inventory undercount is the one that mattered: the section listed five guards where `visit_files` has six callers, and the missing one was `test_wt_spawns_are_pinned` — the caller furthest from the rest, in `tests/common/` rather than `integration_tests/`, and so the likeliest to be re-forked by someone reading that paragraph as the inventory. That's the failure this section exists to prevent, in the section itself. Named now.

On the stray `//!`: removing it as suggested left the module doc butting straight against `use std::fs;`, and the repo's convention is a blank line there — `src/trace/profile.rs`, `src/trace/emit.rs` and `tests/common/mod.rs` all have one. `rustfmt` doesn't manage that position, so nothing would have caught it. Restored in `5844ba094`.

Current state at `5844ba094`: `cargo fmt --check` and `cargo clippy --all-targets -- --deny warnings` clean, 37 tests green across the guard files. Required CI was green on `2b69fb477` before these two doc commits.

Issue #3921, opened 2026-08-26
Plugin marker hooks follow the shell's cwd, so a `cd` mid-session marks the wrong repository


max-sixty:
The six marker hooks in `plugins/worktrunk/hooks/hooks.json` — `marker set` on
`UserPromptSubmit`, `Notification`, `PreToolUse`, `PermissionRequest` and `Stop`,
and `marker clear` on `SessionEnd` — run without `-C`:

```
bash "$CLAUDE_PLUGIN_ROOT/hooks/wt.sh" config state marker set 💬 || true
```

So the marker resolves against the hook process's working directory. That is the
agent's shell cwd, and an agent moves it during a turn — reading files out of a
second checkout, for instance. The two hooks in the same file that do reach for a
directory both scope it: `WorktreeRemove` passes `-C "$p"`, and `WorktreeCreate`
cds to `CLAUDE_PROJECT_DIR`.

Three consequences, all silent:

- A `cd` into another repository marks that repository's branch instead, and the
  session's own marker stops updating. The statusline then carries a stale
  marker on the branch being worked on and a spurious one somewhere else.
- A session that ends while the shell sits in another repository clears that
  repository's marker on `SessionEnd`, and leaves its own set.
- In a directory that is not a repository the command prints
  `✗ git rev-parse --git-common-dir failed (exit 128)` and exits 1. The trailing
  `|| true` turns that into a success, so nothing surfaces.

Repro, given a worktree of one repo and any second repo:

```
cd /tmp/repo-a.feature && wt config state marker set 🧪
# ✓ Set marker for feature to 🧪

cd /tmp/repo-b && wt config state marker set 🧪
# ✓ Set marker for main to 🧪      <- repo-b's branch, from a session on `feature`

cd /tmp/not-a-repo && wt config state marker set 🧪
# ✗ git rev-parse --git-common-dir failed (exit 128), rc 1
```

The marker should track the worktree the session belongs to rather than wherever
its shell currently is. `-C <path>` would carry that, but the value to pass isn't
obvious: `CLAUDE_PROJECT_DIR` is the launch directory and goes stale once the
session moves to another worktree mid-run, and the hook payload's `cwd` is the
same directory that drifted. Pinning the worktree at session start is one route;
resolving it from the session record is another.

Seen on worktrunk e4238e8.

> _This was written by Claude Code on behalf of max-sixty_

max-sixty:
What's the current status of this?

PR #3981, opened 2026-09-01
fix(config): marker set/clear no-op outside a git repository

mahirhir:
## Problem

`marker set`/`marker clear` are called unconditionally by the Claude Code, Codex, and Gemini plugin hooks on nearly every turn (`UserPromptSubmit`, `Notification`, `Stop`, `SessionEnd`, ...), regardless of whether the session's directory is inside a git repository. Outside a repo, `Repository::current()` fails and `handle_state_set`/`handle_state_clear` print

```
✗ git rev-parse --git-common-dir failed (exit 128)
  fatal: not a git repository (or any of the parent directories): .git
```

to stderr and exit 1. The hooks' own `|| true` already discards this, but it's still a spurious failure on every marker call in a session started outside a repo.

This is the third consequence [@worktrunk-bot's triage](https://github.com/max-sixty/worktrunk/issues/3921#issuecomment-5447639733) called out on #3921, and flagged as correct "under every option" independent of how the other two consequences (a shell `cd` retargeting the marker mid-session) get resolved:

> the non-repo `exit 1` swallowed by `|| true` is wrong under every option, since a hook that can't find a repo should be a no-op rather than a suppressed failure.

## Fix

Scoped to the `marker` state key only: when `Repository::current()` fails specifically because the discovery path isn't inside a git repository (checked via `CommandError`'s exit code + stderr, not any other resolution failure like a permission error or a corrupt repo), `handle_state_set`/`handle_state_clear` now return `Ok(())` instead of propagating the error — matching the no-op the hooks already effectively get, minus the stderr noise and the exit 1.

Every other state key (`default-branch`, `previous-branch`, `ci-status`, `logs`) is untouched and keeps failing outside a repo the way every other `wt` command does — `test_state_set_default_branch_outside_repo_still_fails` pins that so the fallback doesn't silently widen later.

I did not touch `hooks.json`/`.codex-plugin/plugin.json`/the Gemini hooks — Claude Code's marker hooks already got `-C "$CLAUDE_PROJECT_DIR"` pinning in #3956 for the *other* two consequences (the `cd`-retargeting ones), and that commit's own message notes Codex and Gemini stay cwd-based "until their hook harnesses expose a stable session project directory." This PR is orthogonal to that: it fixes the swallowed-failure case regardless of whether a given plugin is pinned yet, so it helps Codex/Gemini today and keeps helping them once they get a pin.

## Verification

- New regression tests: `test_state_set_marker_outside_repo_is_noop`, `test_state_clear_marker_outside_repo_is_noop` (exit 0, empty stdout/stderr, run from a tempdir with no `-C`), `test_state_set_default_branch_outside_repo_still_fails` (guards the scoping).
- `cargo test --lib`: 1517 passed, 0 failed.
- `cargo test --test integration`: 2034 passed, 0 failed, 1 ignored (pre-existing, unrelated nightly-only test).
- `cargo clippy --lib --bins -- -D warnings` and `cargo clippy --test integration -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- Manually verified the pre-fix behavior first (`git rev-parse --git-common-dir failed (exit 128)`, exit 1) against current `main`, then the post-fix behavior (exit 0, silent) with the built binary run from a fresh tempdir outside any repo.


mahirhir:
Pushed `525c908`. Took the first of your two options on the stderr match.

**Locale.** `is_missing_repository_error` is now `exit_code == Some(128)` and nothing else. Keeping the substring plus a fragility note would have documented the German user's broken case rather than fixed it, and the suite pins `LC_ALL=C` so nothing would have gone red the day it drifted.

**Docstring.** Rewritten to say what the predicate actually does, including that an unreadable or corrupt `.git` matches too. That is also the honest cost of dropping the substring: any exit 128 out of `git rev-parse --git-common-dir` now takes the no-op arm for `marker`, not just the ones whose stderr says `not a git repository`. For a marker write that still looks like the right call to me, and the key scoping is what keeps it off everything else, but it is a real widening and worth you disagreeing with if you do.

**Tests.** The tempdir and the `current_dir` override are gone from all four new tests, per `wt_command()`'s docstring. `cargo test --test integration -- outside_repo` passes 6 of 6 here, the two pre-existing ones included.

**Doc line.** Reworded the first sentence of the `Don't let a failed marker call fail the session` bullet, and also the clause one bullet above it, which said a hook outside a repository fails. The `|| true` advice stays for the reasons you listed minus the locale one, which this change removes. `skills/` and `plugins/` mirrors carry the same two sentences.

What I could not run: `test_docs_are_in_sync` is `#![cfg(not(windows))]` and this machine is Windows, so I edited the two mirrors by hand instead of letting the sync write them. They are byte-identical to each other and differ from the Starlight source only in the absolute link rewriting that was already there, and neither edited sentence contains a link. If CI disagrees, that is the file to look at.

Also ran: `cargo fmt --check` clean, `cargo clippy --all-targets` with no findings, `cargo test --test integration config_state` 142 passed 0 failed.

mahirhir:
Both of your bounding claims check out independently, so I'll put the evidence here rather than just agree.

**"`marker` is the only key that goes quiet."** Holds on the diff as it stands. The early return is guarded by `key == "marker"` in `handle_state_set` and again in `handle_state_clear`, and `handle_state_get` is untouched, so `default-branch` and every other key still propagate the exit 128 from `Repository::current()`. The guard is on the key rather than on the call site, which is what makes the bound hold for a user typing the command directly and not only for the hooks.

**"The hook path was discarding this through `|| true` anyway."** I counted rather than spot-checked, against `plugins/worktrunk/hooks/hooks.json` on `main`:

```console
$ node count.mjs
hook entries: 8, marker calls: 6, ending in || true: 6
  UserPromptSubmit   guarded
  Notification       guarded
  PreToolUse         guarded
  PermissionRequest  guarded
  Stop               guarded
  SessionEnd         guarded
control (a bare command must read BARE): ok
```

Six of six, no exceptions — so nobody on the shipped plugin was seeing that stderr before this change, and the behaviour they lose is a message they never got. The control line is there because "all six guarded" is the kind of result a broken matcher also returns; a bare command still reads as unguarded.

That does leave the docs bullet doing real work, since `|| true` is a recommendation there and not something the CLI enforces. A user who wrote their own hook without it, outside a repository, went from exit 1 to exit 0 — which is the fix — and a user who wrote their own hook without it in a *corrupt* repository went from a message to silence, which is the widening. Same bound you drew; I just wanted the second half of it written down next to the first.

On the mirrors: agreed, and thanks for tracing which job actually gates it. I had assumed `check-docs` from the name.

PR #3982, opened 2026-09-01
fix(commands): name the configured alias instead of "unrecognized subcommand" outside a repository

yzx9:
`wt co bar` outside a git repository, with `co` a user-config alias, fell through alias dispatch (aliases resolve the current worktree) and PATH lookup to the synthesized `InvalidSubcommand` error — which mixes user-config alias names into its did-you-mean candidates, so the tip suggested the very name the user typed:

```sh
> cat ~/.config/worktrunk/config.toml
[aliases]
co = "wt switch {{ args }}"

> wt co
error: unrecognized subcommand 'co'

  tip: some similar subcommands exist: 'co', 'config'

Usage: wt [OPTIONS] [COMMAND]

For more information, try '--help'.
```

Now `try_alias` itself ends dispatch when no repository is present and the name is a user-config alias:

```sh
> wt co
✗ co is an alias, but there's no git repository here
↳ Run wt inside a repository, or to target one, run wt -C <path> co
```

Only a genuine "not a repository" earns that message: it gates on the structured signal (`git rev-parse` exiting 128), while any other discovery failure — a spawn error from a bad `-C` path, a git that errors for its own reasons — propagates as itself, the way it does for every other command:

```sh
> wt -C /tmp/does-not-exist-xyz co
✗ Failed to execute: git rev-parse --git-common-dir
  No such file or directory (os error 2)
```

The `is_alias` check stays first, so a non-alias name still falls through to its `wt-<name>` PATH binary whatever git did. Because the check lives in alias dispatch, the alias owns its name outside repositories too — a colliding `wt-<name>` PATH binary no longer shadows it there, matching the in-repo precedence ("user config wins over `wt-<name>` PATH binaries"). Inside a repository, dispatch is unchanged.

Also fixed on a sibling path: `similar_subcommands` never suggests the input itself anymore. Non-UTF-8 args skip alias dispatch wholesale, so `wt co $'\xff'` inside a repo still produced `tip: … 'co'` — suggesting the typed name. An exact match is now filtered before `did_you_mean`, which kills the self-suggestion everywhere it can occur; near-match dedupe is unaffected.

Tests: the outside-repo output is pinned by a snapshot (exit code, symbols, styling, hint attachment), with the mock git in `set_git_only_path` answering `rev-parse` the way real git does outside a repository (exit 128 with the fatal message) so the snapshot exercises "no repository" rather than "git failed"; a test proves the alias wins over a colliding PATH binary outside repos; the non-UTF-8 sibling path asserts a tip without `'co'`.

Assisted-by: Claude-Code:GLM-5.3

### DIFF
--- a/src/commands/alias.rs
+++ b/src/commands/alias.rs
@@ -51,10 +51,10 @@ use worktrunk::config::{
     ALIAS_ARGS_KEY, CommandConfig, ProjectConfig, UserConfig, VarScope, alias_context_filter,
     format_alias_variables, referenced_vars_for_config,
 };
-use worktrunk::git::Repository;
+use worktrunk::git::{Repository, WorktrunkError};
 use worktrunk::styling::{
-    eprintln, format_with_gutter, hint_message, info_message, progress_message, verbosity,
-    warning_message,
+    eprintln, error_message, format_with_gutter, hint_message, info_message, progress_message,
+    verbosity, warning_message,
 };
 use worktrunk::trace::Span;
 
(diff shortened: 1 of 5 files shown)


max-sixty:
@worktrunk-bot do you want to finish off these changes, and @yzx9 can object if they dislike them?

yzx9:
@max-sixty @worktrunk-bot Thanks! I’m OK with this patch.

PR #3988, opened 2026-09-02
Scope the branch-inventory staleness contract to membership

max-sixty:
`Repository::local_branches` scans `refs/heads/` once and caches the result for the
lifetime of the `Repository`. Its docstring already scoped a staleness contract to
`commit_sha` — "a snapshot at scan time", resolve through a `RefSnapshot` instead — but
then said of the inventory itself:

> The inventory itself is used for branch listing and upstream-tracking metadata, both of
> which are stable for the duration of a command.

That holds only for a command that runs no hooks. Hooks run *inside* a command, so a
`pre-switch` or `pre-remove` hook that adds, deletes, or renames a branch leaves the
cached membership stale, not just the SHAs.

## Where it shows up

The behaviour is intended — one scan per command is the point — but the docstring said
membership was stable, which sends a reader looking for a bug when they hit it. Two
consumers see it:

- `detect_branch_namespace_conflict` (`src/commands/worktree/switch.rs`) reads the
  inventory to name the branch blocking a `git worktree add -b`. A hook that deletes
  `feature/nested` and creates `feature` in the same command yields "collides with
  existing branch feature/nested" where "already exists" is the truth. Cosmetic.
- `wt step prune` iterates `all_branches()` to choose what to prune, so a branch created
  by a hook mid-command isn't considered. Conservative, but a selection difference rather
  than a message one — which is why the note says membership rather than "messages can be
  wrong".

## What this changes

The docstring, and nothing else. No behaviour change, no new test.

This came out of reviewing #3984, where the same staleness was flagged against the
namespace-conflict mapping. A doc entry was considered and dropped: the FAQ's questions
sit a level up from one cache's staleness, and there's no broader question this is a
detail of.

### DIFF
--- a/src/git/repository/branches.rs
+++ b/src/git/repository/branches.rs
@@ -112,7 +112,9 @@ impl Repository {
     /// needs a current SHA for a ref must resolve through a `RefSnapshot`
     /// captured at the moment of read, not through this inventory. The
     /// inventory itself is used for branch listing and upstream-tracking
-    /// metadata, both of which are stable for the duration of a command.
+    /// metadata, stable for the duration of a command unless that command
+    /// runs a hook: a hook that adds, deletes, or renames a branch leaves
+    /// the membership stale too, not just the SHAs.
     pub fn local_branches(&self) -> anyhow::Result<&[LocalBranch]> {
         Ok(self.local_branch_inventory()?.entries())
     }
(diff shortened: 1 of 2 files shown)


max-sixty:
Folded both in — good catch on the duplicate.

`RepoCache::local_branches` now cross-references `Repository::local_branches` instead of restating the contract, which is the version you suggested: syncing two copies is what let them drift in the first place, and CLAUDE.md's caching pointer routes readers into exactly that field doc.

On the second point — you're right that "membership too, not just the SHAs" reads as enumerating the surface. The accessor now says a hook that touches refs leaves the rest of the inventory stale, "which branches are in it, and where each one's upstream points", so `git branch -u` and a prune that leaves an upstream gone are covered without claiming the list is complete.

Gate green: 4,713 passed, 1 skipped.

PR #4066, opened 2026-09-11
fix: reject removed commits prompt variable

### LINKED ISSUE #2984: Turn deprecated `commits` squash template variable into an error (after 2026-09-06)
## Background

`wt step squash` / `wt merge` build the squash commit-message prompt from a [minijinja template](https://worktrunk.dev/llm-commits/#prompt-templates). Two variables expose the commits being squashed:

- `commits` — a list of bare commit **subject strings** (the original variable).
- `commit_details` — a list of `{ subject, body }` objects (added in #2983 to expose commit bodies).

`commit_details` is a strict superset: `{{ commit.subject }}` reproduces what `commits` gave, and `.body` is newly available. To converge on a single richer variable, we are deprecating `commits`.

## Deprecation cycle

1. **Now (deprecated):** PR #2985 switches the default squash template, docs, and example config to `commit_details` + `.subject`. `commits` still works, but a squash template that references it emits a one-time deprecation warning. Because the shape changes (string → object), the migration is **not** a mechanical rename — a template must change `{{ commit }}` to `{{ commit.subject }}` by hand, so `wt config update` does not auto-rewrite it.
2. **After 2026-09-06 (this issue):** turn a squash template that references `commits` into a hard error (rather than silently rendering an empty/changed value), pointing users at `commit_details`.
3. **Future (separate issue):** once `commits` is fully retired, optionally re-introduce `commits` as the list-of-objects form (aliasing `commit_details`), so the natural name carries the richer data.

## Scope of this issue (step 2)

- [ ] Detect `commits` usage in the rendered squash template and `bail!` with a clear message instead of warning.
- [ ] Update docs/help to drop the deprecated `commits` row (or mark it removed).
- [ ] Add/adjust tests covering the error path.

Filed per maintainer request in https://github.com/max-sixty/worktrunk/pull/2983#issuecomment-4637273025.

fzlzjerry:
Fixes #2984.

## Changes

- Stop supplying the removed `commits` variable to commit/squash prompts and user/project append fragments. Parsed-template validation reports the affected setting, `commit_details`, and `wt config update`; literals, comments, local bindings, and unrelated variables remain unaffected.
- Keep the retirement check and replacement mapping in the config-deprecation layer. Validation runs when building a prompt, not while loading config, so `config show`, `config update`, and unrelated commands can still work with an old config. The existing warning/update invariant and migration rule remain intact.
- Preserve subject-only loops during migration through the existing `CommitDetailValue` rendering. Update the removal/recovery notes and generated help/config documentation. This implements the requested removal step, with no wall-clock check or reintroduction of the old name as an object alias.

## Validation

- `cargo run -- hook pre-merge --yes`: 4,868 tests passed with all features; the existing nightly-only crate-archive build test remains ignored. Pre-commit/clippy, Rust documentation, and doctests passed.
- Regression coverage includes 22 primary/append-template cases, preview failure, real `step squash` and `merge` calls that neither invoke the configured generator nor change branch refs, and a config-update/render/idempotence test.
- An independent old/new binary check confirms that migration preserves the prompt byte-for-byte. Rejected operations leave both refs and the generator call count unchanged; after migration, a real squash invokes that same recorder exactly once and retains the target branch.
- Three targeted omission checks detect removal of either rendering guard or the migration entry; the restored source passes again.
- Generated docs/help synchronization, Astro checks, 15 documentation-renderer tests, and the production docs build passed.

Linux validation used the pinned Rust 1.97 toolchain and Git 2.54.0, plus real bash/zsh/fish/Nushell/PowerShell test shells. The test process used the standard umask and dropped root's file-permission-bypass capabilities so the existing permission assertions were meaningful. The suite's explicit UID-based skips still apply. Native cross-platform CI is still separate from these local results.

### DIFF
--- a/dev/config.example.toml
+++ b/dev/config.example.toml
@@ -358,6 +358,9 @@
 # - `{{ commit_details }}` — list of commits being squashed; each renders as its subject and exposes `.subject` / `.body`
 # - `{{ target_branch }}` — merge target branch
 #
+# The old `commits` variable has been removed. Run `wt config update` to migrate existing prompt templates to `commit_details`.
+# Each detail still renders as its subject when printed directly; `.body` remains available separately.
+#
 # Default template:
 #
 # <!-- DEFAULT_SQUASH_TEMPLATE_START -->
--- a/docs/src/content/docs/config.md
+++ b/docs/src/content/docs/config.md
@@ -479,6 +479,9 @@ Available variables (in addition to commit template variables):
 - `{{ commit_details }}` — list of commits being squashed; each renders as its subject and exposes `.subject` / `.body`
 - `{{ target_branch }}` — merge target branch
 
+The old `commits` variable has been removed. Run `wt config update` to migrate existing prompt templates to `commit_details`.
+Each detail still renders as its subject when printed directly; `.body` remains available separately.
+
 Default template:
 
 <!-- DEFAULT_SQUASH_TEMPLATE_START -->
--- a/docs/src/content/docs/llm-commits.md
+++ b/docs/src/content/docs/llm-commits.md
@@ -181,6 +181,10 @@ Diff:
 | `{{ user_guidance }}` | Rendered user `template-append` fragment (see below) |
 | `{{ project_guidance }}` | Rendered project `template-append` fragment (see below) |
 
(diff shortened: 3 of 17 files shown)


max-sixty:
@worktrunk-bot can you implement this afresh? just remove the deprecated behavior

PR #4132, opened 2026-09-16
fix(flake): use stdenv.hostPlatform.isDarwin in commonArgs

### LINKED ISSUE #4131: flake.nix reads the deprecated stdenv.isDarwin alias, warning on every evaluation
`flake.nix:79` uses `pkgs.stdenv.isDarwin` to gate the `libiconv` build input:

```nix
++ pkgs.lib.optionals pkgs.stdenv.isDarwin [
  libiconv
];
```

nixpkgs defines that name as a deprecated alias — `lib.warn "stdenv.isDarwin is deprecated, use stdenv.hostPlatform.isDarwin instead"` wrapping the real value (`pkgs/stdenv/generic/default.nix`, added 2026-05-09). Reading it emits that warning on every evaluation of this flake, so downstream consumers see it as noise on every `nix develop`, and it becomes a hard error once nixpkgs drops the alias.

`flake.nix:245` already uses `pkgs.stdenv.hostPlatform.system`, as does `nix/home-manager-module.nix`, so this is the one remaining deprecated alias in the repo's Nix files.

`stdenv.hostPlatform.isDarwin` is the identical boolean, so `libiconv` is still selected on Darwin and the build inputs are unchanged.

ashebanow:
Fixes #4131.

`pkgs.stdenv.isDarwin` is a deprecated alias in nixpkgs: it is defined as `lib.warn "stdenv.isDarwin is deprecated, use stdenv.hostPlatform.isDarwin instead"` wrapping the real value (`pkgs/stdenv/generic/default.nix`, added 2026-05-09). Reading it emits that warning on every evaluation of this flake, so downstream consumers see it on every `nix develop`, and it becomes a hard error once nixpkgs drops the alias.

`stdenv.hostPlatform.isDarwin` is the identical boolean, so `libiconv` is still selected on Darwin and the build inputs are unchanged. `flake.nix:245` already uses `pkgs.stdenv.hostPlatform.system`, as does `nix/home-manager-module.nix`, so this makes the file consistent with its own convention.

Verified with `nix flake check` (all outputs evaluate); the warning is gone from an evaluating consumer. No behavior change.


max-sixty:
thanks @ashebanow !

PR #4191, opened 2026-09-19
fix(list): report worktree-side renames

Duang777:
## Summary

- Mark ` R` porcelain records as renames in `wt list`.
- Keep a worktree whose only change is an unstaged rename from rendering as clean.
- Add a direct parser regression test.

Follow-up to the diff-outside observation in #4190. Kept separate so #4190 stayed at its original one-file parser scope.

## Verification

- Red check: the new test fails with the old index-column-only condition.
- `cargo test --bin wt worktree_rename_is_reported_as_dirty`
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- Full suite passed 4,933/4,933 before formatting; the exact formatted rerun passed 4,932 tests and hit the unrelated existing `test_alias_concurrent_receives_sigint` timing failure. Clean CI is the deciding run.


max-sixty:
thanks for these @Duang777 !!

PR #4193, opened 2026-09-20
Fix squash reset atomicity

Bennyjitsu:
## Squash reset atomicity

`wt step squash` / `wt merge --squash`: a commit failure after the soft
reset (e.g. blocked by a pre-commit hook) left the branch's original commits
collapsed into anonymous staged changes instead of restored. Fixed by
capturing the pre-reset SHA and rolling back to it on failure.

Also fixes `has_staged_changes()`, which silently reported "no staged
changes" on a real git failure instead of propagating it — the bug that made
the rollback's own error-check path unreachable.

## Update (this revision)

The PR originally also added an `--author` flag to `merge`/`step commit`/
`step squash`, bundled with this fix since they touched the same functions.
Per this review's feedback: `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` env vars
already propagate through every commit path the flag touched (`git commit`,
`commit-tree` via env, squash), since nothing here scrubs the environment —
verified live against all three paths before removing it. The flag also had
a real correctness bug (its author-spec parser didn't match `git commit
--author`'s own semantics, and the same value was accepted on one merge
path but rejected on another), not worth fixing in code that turned out to
be unnecessary. It's gone; this PR is now just the atomicity fix.

Also fixed: the error-wrapping bug this review flagged in
`restore_after_failed_reset` — the rollback outcome now rides as context on
top of the original error instead of flattening it into a Display string,
so a double failure (commit fails, then the rollback also fails) no longer
loses the original git error from the output. Proven with a direct unit
test (confirmed failing against the prior code, passing against the fix).

Full test suite passes (4,935 tests, 0 failed), full pre-merge gate
(clippy, doc-sync) green.

## Update 2 (this revision)

Fixed a third bug this review flagged: the no-net-changes early return in
`handle_squash` (commits that squash down to zero net diff) left the branch
soft-reset to the merge base while reporting success, instead of restoring
`pre_reset_sha` like the other two post-reset exits already do. `wt merge`'s
cleanup then reads that state as "same commit as target" and deletes the
worktree *and the branch*, discarding the branch's original (unmerged)
commits with no safety backup taken.

This path predates the PR — it wasn't introduced by the atomicity fix above
— but it's the same function, so fixed here rather than filed separately.
`test_merge_squash_empty_changes` now correctly shows the net-zero commits
being merged into the target, instead of the branch being silently deleted
as "already merged".


max-sixty:
I think there is actually something to do here, but it looks quite different to this

